PR #34, opened 2020-04-15
Use $(basename $PWD) to generate code-workspace file

Yajo:
This is helpful if you have more than 1 clone of the same project, maybe one per version or so.

This also removes the need for `_load_answers()` in subprojects, so that code is removed. It can be rescued from git history if needed later again.

@Tecnativa TT20357


pedrobaeza:
Isn't possible to share same code across each version for not duplicating it?

Yajo:
I don't understand... do you mean the test samples?

pedrobaeza:
I see that you have a method `write_code_workspace_file` for each version and I was wondering if they can share a significant amount of code due to DRY principle.

Yajo:
For version v1.x I'm trying to set foundations for future enhancements, and that one could be a good one for the future indeed. However the problem of finding a task manager that works without duplicating code is not easy to solve!

For now I'm focusing on having the same basic features as with doodba-scaffolding. There we also had the `doodbasetup.py` file which [was cloned over and over and did something similar](https://github.com/Tecnativa/doodba-scaffolding/blob/c072cb47665b9fb67288c10270d717e02ced9e4d/.vscode/doodbasetup.py#L76-L94), so this is not worse than that after all.

Also notice that all code inside `tests/default_settings` seems duplicated (in case you're looking to that). That's there so we can easily see what diff produces each change in a new project for each supported Odoo version, but I don't update that manually, I just use `invoke update-test-samples`. I have those samples because this project is a template, and testing it implies being able to review the output it produces.

pedrobaeza:
> For version v1.x I'm trying to set foundations for future enhancements, and that one could be a good one for the future indeed. However the problem of finding a task manager that works without duplicating code is not easy to solve!

OK, but it's something to keep a project mantenible. Imagine when we are at Odoo version 15 and having to update 9 folders or keep some unsynced.

> Also notice that all code inside tests/default_settings seems duplicated (in case you're looking to that). That's there so we can easily see what diff produces each change in a new project for each supported Odoo version, but I don't update that manually, I just use invoke update-test-samples. I have those samples because this project is a template, and testing it implies being able to review the output it produces.

Instead of pre-generating them, isn't better that tests generate them?

Yajo:
> Instead of pre-generating them, isn't better that tests generate them?

Yes, tests generate them and compare them with the pregenerated ones. So we can manually review the pregenerated, and then the test makes sure it works as expected.

If we don't keep them there, how were we going to be able to review results easily? :man_shrugging:

pedrobaeza:
OK then, but if both comes from the same source, it's difficult to hunt down problems as we tend to overpass that, so the efficiency of that test is questionable.

Yajo:
> the efficiency of that test is questionable

The test is doing its job.

For example, look at https://github.com/Tecnativa/doodba-copier-template/pull/34/files#diff-9de5fd74bc4cfbf86147fab9dad9e945

It's checking now that the `doodba.*.code-workspace` file is named after the folder name and not after the project name.

If any future change modifies this, it would fail this test, [as it just did before](https://github.com/Tecnativa/doodba-copier-template/runs/588435376?check_suite_focus=true#step:20:354).

I'll add an additional unit test so maybe in the future we can remove these tests.

pedrobaeza:
OK

Yajo:
Extra test added.

PR #44, opened 2020-04-27
Improve instructions to avoid <pre> problem

Yajo:
Also update submodule.

@Tecnativa TT23674


pedrobaeza:
And how to avoid it in case of CSS?

Yajo:
In the case of CSS it's not a problem. Whitespace never has a meaning in CSS. What you'd have to do is find the xml/html elements affected by those CSS white-space rules and add the ignore tags around them _in the XML files_.

PR #54, opened 2020-05-19
Add some custom downstream tasks

theangryangel:
This PR adds some additional downstream tasks to the copier template that we've found quite useful when migrating from other projects.

The scaffold task also requires a further modification to `devel.yaml.jinja` to make `./odoo/custom` volume non-readonly. I'm not sure if there's a reason for it to be kept readonly, so I've left it out for now, and I'm willing to update/fix the PR if it's helpful.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -183,3 +183,28 @@ def logs(c, tail=10):
         cmd += f" --tail {tail}"
     with c.cd(str(PROJECT_ROOT)):
         c.run(cmd)
+
+
+@task(develop)
+def psql(c, db=None):
+    """Get a psql"""
+    cmd = "docker-compose run --rm odoo psql"
+    if db:
+        cmd += f" {db}"
+    c.run(cmd)
+
+
+@task(develop)
+def shell(c, db=None):
+    """Get an Odoo shell"""
+    cmd = "docker-compose run --rm odoo odoo shell"
+    if db:
+        cmd += f" -d {db}"
+    c.run(cmd)
+
+
+@task(develop)
+def scaffold(c, name):
+    """Create a scaffold"""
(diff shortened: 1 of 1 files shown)


tasks_downstream.py
```diff
@@ -183,3 +183,28 @@ def logs(c, tail=10):
...
+    cmd = "docker-compose run --rm odoo odoo shell"
+    if db:
+        cmd += f" -d {db}"
+    c.run(cmd)
+
+
+@task(develop)
+def scaffold(c, name):
+    """Create a scaffold"""
+    cmd = f"docker-compose run --rm odoo odoo scaffold {name} /opt/odoo/custom/src/private"
```
Yajo, line 209: this is mounted as readonly because that's how it will be in production.

Maybe you can just add a `-v` to this `docker-compose run`?

Also TBH I personally will not use this. I find https://github.com/OCA/maintainer-tools/tree/master/template/module to be much more complete and I just copy-paste it. A proper copier template for a module would be best IMHO.

tasks_downstream.py
```diff
@@ -183,3 +183,28 @@ def logs(c, tail=10):
...
+    cmd = "docker-compose run --rm odoo psql"
+    if db:
+        cmd += f" {db}"
+    c.run(cmd)
+
+
+@task(develop)
+def shell(c, db=None):
+    """Get an Odoo shell"""
+    cmd = "docker-compose run --rm odoo odoo shell"
```
Yajo, line 200:
```suggestion
    cmd = "docker-compose run --rm odoo click-odoo"
```

This provides a shell out of the box for v8 too.

Yajo:
Needs tests.

theangryangel:
I've discovered some issues with these proposed changes. 

Rather than leave this PR open for a while (it's going to take a week or two until I get around to it), I'll open a new one when I get to it :)

PR #72, opened 2020-06-18
Github Actions: don't run heavy tests on doc updates

yelizariev:
Docs: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#excluding-paths

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,8 +2,16 @@ name: test
 
 on:
   pull_request:
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/**.md
   push:
     branches: [master]
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/**.md
 
 jobs:
   test:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,10 +45,6 @@ jobs:
         with:
           path: .venv
           key: venv ${{ env.PY }}
-      - uses: actions/cache@v1
-        with:
-          path: ~/.cache
-          key: pre-commit ${{ env.PY }} ${{ hashFiles('.pre-commit-config.yaml') }}
       - run: pip install pipx
       - name: Generate cache key PIPX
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -10,6 +10,10 @@ on:
       - synchronize
       - unlabeled
       - unlocked
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/**
```
Yajo, line 16: Revert this please. If I'm not wrong, I might still be able to use automerge for doc updates.

.github/workflows/test.yml
```diff
@@ -2,8 +2,16 @@ name: test
...
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/**
   push:
     branches: [master]
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/**
```
Yajo, line 14: This workflow runs both pre-commit and tests.

Tests are not so important, but pre-commit is for docs.

Also there are tests that test the correct behavior of `docs/scaffolding2copier.sh`, so I think it would be better to use `docs/**.md` instead.

yelizariev: Not sure that I understand you about pre-commit. From my understanding pre-commit runs in ``lint.yml``. 

Take a look on new updates, please

Yajo: You're right, I was confused due to some garbage I seem to have copy-pasted from `lint.yml`. I fixed it myself.

.github/workflows/test.yml
```diff
@@ -2,8 +2,16 @@ name: test
 
 on:
   pull_request:
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/*.md
```
Yajo, line 8:
```suggestion
      - docs/**.md
```

.github/workflows/test.yml
```diff
@@ -2,8 +2,16 @@ name: test
...
   push:
     branches: [master]
+    paths-ignore:
+      - README.md
+      - CONTRIBUTING.md
+      - docs/*.md
```
Yajo, line 14:
```suggestion
      - docs/**.md
```

PR #81, opened 2020-08-01
mount ssh for `invoke git-aggregate` to deal with private repos, see #80 for details, closes #80

### LINKED ISSUE #80: how project's `odoo/custom/ssh` is supposed to propogate inside the container if ssh folder is not mounted as a volume in setup-devel.yaml?
https://github.com/Tecnativa/doodba-copier-template/blob/9a33b2e3adf4f98b2f93f02ee99189271f8a5461/setup-devel.yaml.jinja#L32

iledarn:


Yajo:
See https://github.com/Tecnativa/doodba-copier-template/issues/80#issuecomment-667856536.

PR #124, opened 2020-09-29
Ask odoo version first

yelizariev:
It allows to quickly apply default values on all other options by pressing Enter button.
Such approach is especially useful for local modules development

----------------
I'd actually like to move just odoo_version attribute, but the attributes are grouped, so I moved whole ``odoo_`` group and reordered making first the attributes most important for modules development.

As for now, I need to slowly click enter to don't miss odoo_version question -- the only question where I don't use default value

------------------

I just noticed that it's stable branch. Am I supposed to send it to devel branch instead?

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -77,6 +77,75 @@ _migrations:
         - "{{ _copier_conf.dst_path }}"
 
 # Questions for the user
+odoo_version:
+  help: On which odoo version is it based?
+  type: float
+  default: 13.0
+  choices:
+    - 7.0
+    - 8.0
+    - 9.0
+    - 10.0
+    - 11.0
+    - 12.0
+    - 13.0
+
+odoo_dbfilter:
+  default: ".*"
+  type: str
+  help: >-
+    Set your Odoo db filter. It must be a regexp that matches the domain name being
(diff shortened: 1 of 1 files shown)


copier.yml
```diff
@@ -77,6 +77,75 @@ _migrations:
...
+    Examples: en_US, es_ES, es_VE...
+
+odoo_admin_password:
+  secret: true
+  default: example-admin-password
+  type: str
+  help: >-
+    💡 To auto-generate strong passwords, see https://ddg.gg/?q=password+64+strong
+
+    ⚠️ This password is critical for security, especially if you have set odoo_listdb to
```
joao-p-marques, line 128: As we are changing the order, maybe change this to 
```suggestion
    ⚠️ This password is critical for security, especially if you set odoo_listdb to
```
or put `odoo_listdb` before this.

joao-p-marques (commented):
LGTM :+1: 
Just a small comment

Yajo (approved):

copier.yml
yelizariev: Yeah, makes sense

joao-p-marques (approved):
LGTM :+1:

PR #127, opened 2020-09-30
Update ptvsd -> debugpy

joao-p-marques:
Seems like debugpy is the recommended library for debugging Python with VSCode, so it makes sense to make this transition.

https://code.visualstudio.com/docs/python/debugging#_debugging-by-attaching-over-a-network-connection 
https://github.com/microsoft/debugpy

This updates the template, but projects Doodba needs to be updated (see https://github.com/Tecnativa/doodba/pull/324)

Tecnativa-Task #25875

### DIFF
--- a/devel.yaml.jinja
+++ b/devel.yaml.jinja
@@ -34,7 +34,7 @@ services:
     environment:
       DOODBA_ENVIRONMENT: "${DOODBA_ENVIRONMENT-devel}"
       LIST_DB: "true"
-      PTVSD_ENABLE: "${DOODBA_PTVSD_ENABLE:-0}"
+      DEBUGPY_ENABLE: "${DOODBA_DEBUGPY_ENABLE:-0}"
       PGDATABASE: &dbname devel
       PYTHONDONTWRITEBYTECODE: 1
       PYTHONOPTIMIZE: ""
--- a/tasks.py.jinja
+++ b/tasks.py.jinja
@@ -56,7 +56,7 @@ def write_code_workspace_file(c, cw_path=None):
     except (FileNotFoundError, json.decoder.JSONDecodeError):
         pass  # Nevermind, we start with a new config
     # Launch configurations
-    ptvsd_configuration = {
+    debugpy_configuration = {
         "name": "Attach Python debugger to running container",
         "type": "python",
         "request": "attach",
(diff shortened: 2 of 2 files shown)


tasks.py.jinja
```diff
@@ -277,13 +277,13 @@ def lint(c, verbose=False):
...
 @task(develop)
-def start(c, detach=True, ptvsd=False):
+def start(c, detach=True, debugpy=False):
     """Start environment."""
     cmd = "docker-compose up"
     if detach:
         cmd += " --detach"
     with c.cd(str(PROJECT_ROOT)):
-        c.run(cmd, env=dict(UID_ENV, DOODBA_PTVSD_ENABLE=str(int(ptvsd))))
+        c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
```
Yajo, line 286: Do you think we could use this opportunity to add also support for `DEBUGPY_ARGS`?

I see an opportunity for that. It'd be nice if Odoo was started within debugpy always when running in vscode, and then starting the debug session would be more instant.

joao-p-marques: Yes, we could do that. I have seen a comment on the code, however, stating it would also make Odoo slower... Can that make a difference?

devel.yaml.jinja
```diff
@@ -10,7 +10,7 @@ services:
       default:
       public:
     ports:
-      - "127.0.0.1:{{ macros.version_major(odoo_version) }}899:6899"
+      - "127.0.0.1:{{ macros.version_major(odoo_version) }}678:5678"
```
Yajo, line 13: Is there a real need for this? Because otherwise I prefer to just use port 6899 like until today. It's used by `pudb` in remote mode also, if you want to use that debugger.

tasks.py.jinja
Yajo: Yes, it's always slower, but having the option can help.

Depending on what you're doing, you need more speed or you need better debugability.

In any case, it can be done later if you prefer.

devel.yaml.jinja
joao-p-marques: Ok. I removed the last commit with the changes

tasks.py.jinja
joao-p-marques: Well, in that case I was thinking maybe for another time. It is not critical and it would speed up adding debugpy, which is more important.

Yajo (approved):

PR #139, opened 2020-10-13
Fix debugger timeout

joao-p-marques:
Add timeout to invoke start task for debugging sessions
   
Add option to pass --sleep argument to inv start with number of seconds
to sleep.
Use that argument in VSCode compound definition.
See https://github.com/microsoft/vscode-python/issues/14149
    
TT26189

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -69,7 +69,7 @@ def lint(c, verbose=False):
 @task(develop)
 def test(c, verbose=False):
     """Test project."""
-    flags = ["-n", "auto", "--color=yes"]
+    flags = ["--color=yes"]
     if verbose:
         flags.append("-vv")
     flags = " ".join(flags)
--- a/tasks.py
+++ b/tasks.py
@@ -69,7 +69,7 @@ def lint(c, verbose=False):
 @task(develop)
 def test(c, verbose=False):
     """Test project."""
-    flags = ["--color=yes"]
+    flags = ["-n", "auto", "--dist", "loadscope", "--color=yes"]
     if verbose:
         flags.append("-vv")
     flags = " ".join(flags)
--- a/tasks.py
+++ b/tasks.py
@@ -67,14 +67,22 @@ def lint(c, verbose=False):
 
 
 @task(develop)
-def test(c, verbose=False):
-    """Test project."""
-    flags = ["-n", "auto", "--dist", "loadscope", "--color=yes"]
+def test(c, verbose=False, sequential=False):
+    """Test project.
+
+    Add --sequential to run only sequential tests, with parallelization disabled.
+    """
+    flags = ["--color=yes"]
     if verbose:
         flags.append("-vv")
+    if sequential:
+        flags.extend(["-m", "sequential"])
+    else:
+        flags.extend(["-n", "auto", "-m", '"not sequential"'])
     flags = " ".join(flags)
+    cmd = f"poetry run pytest {flags} tests"
     with c.cd(str(TEMPLATE_ROOT)):
-        c.run(f"poetry run pytest {flags} tests")
+        c.run(cmd)
 
 
 @task(develop)
--- a/tasks.py.jinja
+++ b/tasks.py.jinja
@@ -9,6 +9,9 @@ Contains common helpers to develop using this child project.
 """
 import json
 import os
+import time
+import subprocess
+from logging import getLogger
 from itertools import chain
 from pathlib import Path
 from shutil import which
@@ -20,6 +23,7 @@ PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
 UID_ENV = {"GID": str(os.getgid()), "UID": str(os.getuid()), "UMASK": "27"}
 
+_logger = getLogger(__name__)
 
 @task
 def write_code_workspace_file(c, cw_path=None):
@@ -181,7 +185,7 @@ def write_code_workspace_file(c, cw_path=None):
                 "label": "Start Odoo in debug mode",
                 "type": "process",
                 "command": "invoke",
-                "args": ["start", "--detach", "--debugpy"],
+                "args": ["start", "--detach", "--debugpy", "--sleep", "10"],
                 "presentation": {
                     "echo": True,
                     "reveal": "silent",
@@ -294,13 +298,27 @@ def lint(c, verbose=False):
 
 
 @task(develop)
-def start(c, detach=True, debugpy=False):
+def start(c, detach=True, debugpy=False, wait=None):
     """Start environment."""
     cmd = "docker-compose up"
     if detach:
         cmd += " --detach"
     with c.cd(str(PROJECT_ROOT)):
         c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
+    if wait:
+        _logger.info("Waiting for services to spin up...")
+        curl_stdout = None
+        # Odoo
+        while(not curl_stdout):
+            curl_cmd = "curl 127.0.0.1:12069"
+            curl_stdout = subprocess.run(curl_cmd.split(' '), stdout=subprocess.PIPE, stderr=subprocess.PIPE).stdout
+        _logger.info("Odoo started.")
+        # Debugpy
+        if debugpy:
+            while(not curl_stdout or "(52) Empty reply from server" in curl_stdout):
+                curl_cmd = "curl 127.0.0.1:12899"
+                curl_stdout = subprocess.run(curl_cmd.split(' '), stdout=subprocess.PIPE, stderr=subprocess.PIPE).stdout
+            _logger.info("Debugpy started.")
 
 
 @task(
--- a/tasks.py.jinja
+++ b/tasks.py.jinja
@@ -10,7 +10,6 @@ Contains common helpers to develop using this child project.
 import json
 import os
 import time
-import subprocess
 from logging import getLogger
 from itertools import chain
 from pathlib import Path
@@ -22,8 +21,10 @@ ODOO_VERSION = {{ odoo_version }}
 PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
 UID_ENV = {"GID": str(os.getgid()), "UID": str(os.getuid()), "UMASK": "27"}
+SERVICES_WAIT_TIME = int(os.environ.get("SERVICES_WAIT_TIME", 3))
 
 _logger = getLogger(__name__)
+
 
 @task
 def write_code_workspace_file(c, cw_path=None):
@@ -185,7 +186,7 @@ def write_code_workspace_file(c, cw_path=None):
                 "label": "Start Odoo in debug mode",
                 "type": "process",
                 "command": "invoke",
-                "args": ["start", "--detach", "--debugpy", "--sleep", "10"],
+                "args": ["start", "--detach", "--debugpy"],
                 "presentation": {
                     "echo": True,
                     "reveal": "silent",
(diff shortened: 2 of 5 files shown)


joao-p-marques:
I've just noticed I have also included another commit by accident (6ee4f5c) which was on `devel` but not on `stable`.

Should I leave this as is, or instead make the PR targeting `devel`?

joao-p-marques:
Need to add some tests to this

tasks.py.jinja
```diff
@@ -283,13 +284,15 @@ def lint(c, verbose=False):
...
-def start(c, detach=True, debugpy=False):
+def start(c, detach=True, debugpy=False, sleep=0):
     """Start environment."""
     cmd = "docker-compose up"
     if detach:
         cmd += " --detach"
     with c.cd(str(PROJECT_ROOT)):
         c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
+    if sleep:
+        time.sleep(sleep)
```
Yajo, line 295: This is usually not a good idea, because you will slow down fast machines, or you will have the same bug in slow machines.

It's better to query the port and unblock the task when it responds as listening.

joao-p-marques: Ok. Thanks for the suggestion. I will work on that.

joao-p-marques:
Tests are failing due to port conflicts when trying to do `invoke start` in the test. Can you take a look @Yajo ?

Yajo:
This seems a problem of parallel tests that collide in opening the same port, due to how we use https://github.com/pytest-dev/pytest-xdist

Try removing `-n auto` from here while you're in draft. https://github.com/Tecnativa/doodba-copier-template/blob/4b85e88858b2e56aa1be25adacdd5d9953897198/tasks.py#L72

If you make pass the tests sequentially, we'll handle the parallel problem later.

Marking as draft, as I see the code you have is mostly that.

tasks.py.jinja
```diff
@@ -283,13 +286,27 @@ def lint(c, verbose=False):
...
 @task(develop)
-def start(c, detach=True, debugpy=False):
+def start(c, detach=True, debugpy=False, wait=None):
     """Start environment."""
     cmd = "docker-compose up"
     if detach:
         cmd += " --detach"
     with c.cd(str(PROJECT_ROOT)):
         c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
+    if wait:
```
gdgellatly, line 296: if it helps, a snippet of a function I use to wait for Pycharm debug to be open. Could be adjusted to work in a while loop.

```python
    import socket

    def socket_is_open(host, port):
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        if sock.connect_ex((host, port)) == 0:
            return True
        return False
```

joao-p-marques:
Thanks for the suggestion @gdgellatly 
I am using it in the tests.
However, when I tried to use it to actually monitor if the service is up or not, I get some errors, mainly because we are actually proxy-ing that port through another container, so the socket will always connect. 
Also, it seems that debugpy is a bit delicate in that regard, and it is possible that, after receiving the connection attempt from the socket, it doesn't receive the attach request from the actual debugger, and that can be the reason behind the errors I was getting. 

I ended up trying another approach: I monitor `debugpy` logs (we need to had that option in Doodba, I was just testing locally) through a tmp file in the container. I don't know if it will have a heavy impact on performance, but I don't think it will be a major problem. 

What do you think @Yajo?

gdgellatly:
@joao-p-marques FWIW, There is actually a way you can access a debug port on a private isolated network in docker without having to proxy the port.  Even docker private networks have a gateway and this corresponds to localhost and is always the first ip in the private network address range. If you listen on that IP for debugger rather than 127.0.0.1 or localhost it magaically works, but only for localhost debugger connections. (Basically you use the output of this function as IP in settrace.)

It is something I got working but never properly as it assumes /24 subnets but here is that snippet run from inside container. The `10.x.x.x.` IP address is just a dummy address, the socket returns the container IP and we just replace with a 1. I imagine there is a way to get the gateway IP better than a string replace.

The main benefit is you can have multiple Odoo sessions debugging without having to do any port mapping or the like to avoid listen conflicts.

```python
    def get_ip():
        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
        try:
            s.connect(('10.255.255.255', 1))
            IP = s.getsockname()[0]
        except Exception:
            IP = '127.0.0.1'
        finally:
            s.close()
        parts = list(IP.rpartition("."))
        parts[-1] = "1"
        return "".join(parts)
```

Yajo:
@joao-p-marques Can't debugpy log to stdout so you can monitor it easier?

@gdgellatly Can't you get the IP by inspecting the container? Because since v2.0.0 we support having 1 doodba devel environment per odoo version up simultaneously. If you just pick up the 1st IP blindly, I guess you are prone to start debugging the wrong container easily.

gdgellatly:
> @joao-p-marques Can't debugpy log to stdout so you can monitor it easier?
> 
> @gdgellatly Can't you get the IP by inspecting the container? Because since v2.0.0 we support having 1 doodba devel environment per odoo version up simultaneously. If you just pick up the 1st IP blindly, I guess you are prone to start debugging the wrong container easily.

@Yajo you are thinking in reverse. In this case it is not the localhost needing to get the IP of the container, it is the container having to get the gateway IP of the localhost, so you can't docker inspect from inside. If they are listening on different ports it doesn't matter, the IP will always be the same. In any case, it is just an idea to avoid the socat proxy. THe use case this came out of was actually connecting to a remote debug server so the case is maybe slightly different, although underneath it looks the same to me. 

Maybe you are right and what we really want is the listen interface of the vscode debugger to be network interface of the network. THere is nothing to stop you port mapping directly on container if there are multiple ones.

joao-p-marques:
> @joao-p-marques Can't debugpy log to stdout so you can monitor it easier?

Yes, it can. The problem is that it outputs a lot of information, both from the debugpy server and the adapter handling the client, so it would end up flooding our container logs and make it a lot harder to debug Odoo itself. 

Given that, I think it is better to keep it isolated.

@gdgellatly Thanks for you suggestion. If this ends up not being a good solution, I will further investigate that.

tasks.py.jinja
```diff
@@ -283,14 +286,20 @@ def lint(c, verbose=False):
...
         cmd += " --detach"
     with c.cd(str(PROJECT_ROOT)):
         c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
-
+    if wait:
+        _logger.info("Waiting for services to spin up...")
+        debugpy_log = None
+        while(not debugpy_log or b"Adapter is accepting incoming client connections" not in debugpy_log):
+            cmd = "docker-compose exec -T odoo cat /tmp/debugpy-log/debugpy.server-1.log"
+            debugpy_log = subprocess.run(cmd.split(' '), stdout=subprocess.PIPE, stderr=subprocess.PIPE).stdout
```
Yajo, line 301: Use `c` instead of `subprocess`, to be consistent. Invoke provides it, with a handful of niceties.

Yajo:
OK I understand your point now @gdgellatly. I think it's better to avoid such low-level plumbing if possible. Right now the proxy adds almost zero overhead AFAICS (I didn't measure that, but local-to-local connections shouldn't be a big worry) so that's not a real "problem" we're trying to solve.

@joao-p-marques your solution then looks fine. I guess the best you can do is to then use a log file under `/opt/odoo/auto` in the container, which maps to `./odoo/auto` outside of it and it's our supported file interchange gateway: https://github.com/Tecnativa/doodba-copier-template/blob/4b85e88858b2e56aa1be25adacdd5d9953897198/devel.yaml.jinja#L48

joao-p-marques:
Changes made @Yajo 
I also opened a PR on Doodba to enable the debugpy logs to output to a file.

gdgellatly:
> OK I understand your point now @gdgellatly. I think it's better to avoid such low-level plumbing if possible. Right now the proxy adds almost zero overhead AFAICS (I didn't measure that, but local-to-local connections shouldn't be a big worry) so that's not a real "problem" we're trying to solve.
> 
> @joao-p-marques your solution then looks fine. I guess the best you can do is to then use a log file under `/opt/odoo/auto` in the container, which maps to `./odoo/auto` outside of it and it's our supported file interchange gateway:
> 
> https://github.com/Tecnativa/doodba-copier-template/blob/4b85e88858b2e56aa1be25adacdd5d9953897198/devel.yaml.jinja#L48

@Yajo the proxy adds immaterial latency overhead but it severely throttles bandwidth. transfers traversing a socat proxy from/to localhost to container and vice versa drop from essentially wirespeed to around 1MB/s.

gdgellatly:
@yajo @joao-p-marques is not a simpler solution to this to just remove the --wait-for-client flag and then just put a sleep in the task. I mean that should avoid all startup delays. In Pycharm when I integrated the debugger we never waited for the client. It has the added benefit of actually just being able to run full time in dev. It also means with a little care you can enable debug quite safely in test servers.

--wait-for-client seems a bit superfluous for the Odoo use case which is a long running app. Very rarely are we wanting to debug the initial startup code.

joao-p-marques:
Well, I actually ran into some issues when removing the `--wait-for-client` flag. The debugger in VSCode couldn't attach properly. I can test more extensively, though.

The reason we opted no to go with a simple "sleep" on the task was this: https://github.com/Tecnativa/doodba-copier-template/pull/139#discussion_r504455509 
It could potentially be inconsistent and generate errors.

gdgellatly:
> Well, I actually ran into some issues when removing the `--wait-for-client` flag. The debugger in VSCode couldn't attach properly. I can test more extensively, though.
> 

Yes, it will cause problems if you don't add a sleep. Remember the issue here is not that the server is waiting for the client, it is that the client is timing out waiting for the server. 

> The reason we opted no to go with a simple "sleep" on the task was this: [#139 (comment)](https://github.com/Tecnativa/doodba-copier-template/pull/139#discussion_r504455509)
> It could potentially be inconsistent and generate errors.

Using sleep won't generate any errors. I just have a constant in my tasks.py to tune it on start. On my laptop, that requires a value of 0.75 on my desktop it needs 1.25. Those are probably conservative, but even just defaulting a value of 2 seconds to be safe, I doubt anyone cares about 2 seconds of delay given overall odoo startup time is over 2 seconds anyway, especially if remving wait for client, you could go 5 and noone would notice.  It is really only about 1 second, my laptop is pretty fast and needs 0.75 already, my desktop which is 6 years old, Windows, 4GB and only an i5 chip gets away with 1.25, so I doubt it actually makes that much difference machine speed.

joao-p-marques:
Yes, maybe you are right.
WDYT @Yajo?

Yajo:
After seeing how this is becoming so hard, let's go for the simpler-but-less-perfect solution. Add a default that seems to work (2, 3s maybe), but let users configure it with an env variable.

Also if we can drop the `--wait-for-connect`, it'd be good too.

Besides, you might consider checking https://github.com/firefox-devtools/vscode-firefox-debug/issues/227#issuecomment-706716961 to improve the firefox debugging experience too (BTW), but this can be a separate issue if you prefer.

joao-p-marques:
Tests will fail until https://github.com/Tecnativa/doodba/pull/334 is merged.

joao-p-marques:
I think this is already ready to merge. 
We just have the problem you mentioned in https://github.com/Tecnativa/doodba-copier-template/pull/139#issuecomment-709909503
As I have added tests for the start task, we are no longer running them in parallel, which will make them slower...

devel.yaml.jinja
```diff
@@ -35,6 +35,7 @@ services:
       DOODBA_ENVIRONMENT: "${DOODBA_ENVIRONMENT-devel}"
       LIST_DB: "true"
       DEBUGPY_ENABLE: "${DOODBA_DEBUGPY_ENABLE:-0}"
+      DEBUGPY_ARGS: "--listen 0.0.0.0:6899 --wait-for-client --log-to /tmp/debugpy-logs/"
```
Yajo, line 38: This is not necessary anymore, right?

tasks.py.jinja
```diff
@@ -283,14 +288,16 @@ def lint(c, verbose=False):
 
 
 @task(develop)
-def start(c, detach=True, debugpy=False):
+def start(c, detach=True, debugpy=False, wait=False):
```
Yajo, line 291: I don't think you need this argument.

Doing `env SERVICES_WAIT_TIME=0 inv start` would be the same, after all...

Yajo (changes requested):

tasks.py.jinja
joao-p-marques: Changes made

joao-p-marques:
@Yajo 
Seems to be working with your suggestion of using `--dist loadscope`.
Tests might be taking a couple more minutes, but I think it is acceptable, no?

tasks.py
```diff
@@ -67,14 +67,19 @@ def lint(c, verbose=False):
...
-def test(c, verbose=False):
+def test(c, verbose=False, concurrent=False):
     """Test project."""
     flags = ["-n", "auto", "--color=yes"]
     if verbose:
         flags.append("-vv")
+    if concurrent:
+        flags.extend(["--dist", "loadscope", "-m", "concurrent"])
+    else:
+        flags.extend(["-m", '"not concurrent"'])
```
Yajo, line 78:
```suggestion
def test(c, verbose=False, sequential=False):
    """Test project.

    Add --sequential to run only sequential tests, with parallelization disabled.
    """
    flags = ["--color=yes"]
    if verbose:
        flags.append("-vv")
    if concurrent:
        flags.extend(["-m", "sequential"])
    else:
        flags.extend(["-n", "auto", "-m", '"not sequential"'])
```

tests/test_tasks_downstream.py
```diff
@@ -88,3 +89,52 @@ def test_resetdb(
...
+            invoke("stop")
+            stdout = invoke("start", "--debugpy")
+            assert socket_is_open("127.0.0.1", int(supported_odoo_version) * 1000 + 899)
+    finally:
+        # Imagine the user is in the odoo subrepo for this command
+        with local.cwd(tmp_path / "odoo" / "custom" / "src" / "odoo"):
+            invoke("stop", "--purge")
+
+
+def socket_is_open(host, port):
```
Yajo, line 136: Move helpers to the top if they are only important for one test file, and to `conftest.py` if they could be potentially reused.

Yajo (changes requested):

joao-p-marques:
Changes made @Yajo 
Thak you!

Yajo (approved):

Yajo:
A long path towards a small patch, after all! But we learned a lot in the path. Thanks @joao-p-marques for your code and patience.

joao-p-marques:
Haha thank you! I am going to integrate this in #160

PR #141, opened 2020-10-15
Add script to proxy pycharm debugger port into odoo container

ap-wtioit:
Pycharm expects its dynamic debugger port to be reachable on 127.0.0.1
 on the host system. This adds a script that watches for pydevd debugger
 ports and updates a proxy service accordingly


Yajo:
Humm... :confused: Some points:

1. I don't use pycharm, and we're currently on a transition to try to get all our workmates that still use it to leave it, so I'm not sure I want to maintain this... :thinking: 
2. Can't pycharm use a port such as this one, which is used for all the other debuggers? https://github.com/Tecnativa/doodba-copier-template/blob/4b85e88858b2e56aa1be25adacdd5d9953897198/devel.yaml.jinja#L14
3. If both points are solved, then the script should be an invoke task instead.

ap-wtioit:
i don't see how 1 could be resolved by me so 3 can also not be solved by me.

Leaves number 2:
No Pycharm can't (or not yet for more than 1,5 years). It randomizes the port and chooses an unused one, which is the reason for this script. I developed it after a short discussion on https://discord.com/channels/737652535149592587/765874032192520202 today for the OCA days. I always workarrounded the problem by using `networks: *public` on the odoo service in devel.yaml which disables the network isolation and was pointed out as not ideal by others (and i agree).

So i wrote this short script. As the others also used doodba with pycharm i thought this would be the right place to share it with them as this is the base of our common Odoo development.

unfortunately the whole copier and commit hook stuff doesn't work on my ubuntu 18.04 LTS as npm reports an error somwhere deep in the dependencies. so i have no real way currently of creating an invoke task. but if you decide supporting pycharm would be ok then i could investigate what's up with the npm stuff.

Yajo:
I have no idea about pycharm, but https://www.jetbrains.com/help/pycharm/run-debug-configuration-python-remote-debug.html says you can specify a port...

Yajo:
> unfortunately the whole copier and commit hook stuff doesn't work on my ubuntu 18.04 LTS as npm reports an error somwhere deep in the dependencies

Open a new issue about the subject and we can discuss that, of course! :wink: 

> if you decide supporting pycharm

TBH that's not going to happen... :man_shrugging: 

However, remember that with Copier you can have several templates applied to the same project. Check out https://copier.readthedocs.io/en/stable/configuring/#the-copier-answersyml-file 

So, you can use this template to reuse all of our work, and have a separate template (tracked, say, in `.copier-answers.pycharm.yml`) that applies your customizations. I'll be happy to provide a link to that subtemplate here in case you do that, for those that want pycharm support.

Thus closing, as I think this is not gonna be merged anyway. I hope you understand, and thanks for your contribution. :blush:

PR #149, opened 2020-10-21
Migrate prettier from prettier/prettier to prettier/pre-commit

joao-p-marques:
See https://github.com/prettier/prettier/issues/9459

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,8 +33,8 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies: [flake8-bugbear==20.1.4]
-  - repo: https://github.com/prettier/prettier
-    rev: 2.1.1
+  - repo: https://github.com/prettier/pre-commit
+    rev: v2.1.1
     hooks:
       - id: prettier
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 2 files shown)


joao-p-marques:
Problem with Still depending on https://github.com/prettier/pre-commit/issues/16 fixed

joao-p-marques:
When https://github.com/prettier/pre-commit/pull/17 and https://github.com/prettier/pre-commit/pull/18 are merged, we can remove the extra dependencies and args.

.vscode/settings.json
```diff
@@ -32,7 +32,6 @@
   "python.formatting.provider": "black",
   "python.linting.flake8Enabled": true,
   "python.linting.pylintEnabled": true,
-  "python.pythonPath": "${workspaceRoot}/.venv/bin/python",
```
Yajo, line 35: I guess you should leave this here.

Yajo (changes requested):

.vscode/settings.json
joao-p-marques: Sorry, VSCode keeps making this change on it's own and I committed it by accident.

joao-p-marques: Changes made

Yajo (approved):

PR #151, opened 2020-10-22
Add higher priority traefik flag to Odoo routers

joao-p-marques:
Add a fixed priority to Odoo Traefik routers.

This helps if we want to setup a second service that uses the same or a CatchAll rule and use it only when Odoo fails.

TT25801

### DIFF
--- a/_traefik2_labels.yml.jinja
+++ b/_traefik2_labels.yml.jinja
@@ -72,13 +72,19 @@
         {{ domain_group.cert_resolver }}
       {%- endif %}
 
+      {#- Give higher priority #}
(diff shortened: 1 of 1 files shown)


Yajo (changes requested):
Tests fail, as you can see.
Also please include a description in the commit message more focused on why you are doing this, not what you are doing (which is already seen in the code itself :wink:).

joao-p-marques:
Yes, sorry, MB.

I am trying to fix the tests with the new pushed changes.

joao-p-marques:
This is no longer needed because the default priority set up by Traefik is already enough for us.

PR #150, opened 2020-10-22
Only create postgres_db if not using external

joao-p-marques:
Fixes https://github.com/Tecnativa/doodba-copier-template/issues/148

PR #155, opened 2020-10-23
Add newline between db and previous service

Yajo:
Tiny regression from #150. I like it better as it was before (extra newline among services) for readability.


joao-p-marques (approved):
LGTM :+1:

PR #159, opened 2020-10-28
Odoo only depends on db if not using external db

joao-p-marques:
Odoo can't always depend on `db` because we don't always have it. This fixes that bug.

Fixes https://github.com/Tecnativa/doodba-copier-template/pull/150#issuecomment-717671877

### DIFF
--- a/devel.yaml.jinja
+++ b/devel.yaml.jinja
@@ -48,7 +48,9 @@ services:
       - ./odoo/auto:/opt/odoo/auto:rw,z
     depends_on:
       - cdnjs_cloudflare_proxy
+      {% if postgres_version -%}
       - db
+      {%- endif %}
       - fonts_googleapis_proxy
       - fonts_gstatic_proxy
       - google_proxy
--- a/prod.yaml.jinja
+++ b/prod.yaml.jinja
@@ -20,7 +20,9 @@ services:
       SMTP_SERVER: smtplocal
       {%- endif %}
     depends_on:
+      {% if postgres_version -%}
       - db
+      {%- endif %}
       {%- if smtp_relay_host %}
       - smtp
       {%- endif %}
--- a/test.yaml.jinja
+++ b/test.yaml.jinja
@@ -20,7 +20,9 @@ services:
       SMTP_SERVER: smtplocal
     restart: unless-stopped
     depends_on:
+      {% if postgres_version -%}
       - db
+      {%- endif %}
       - smtp
     networks:
       default:


devel.yaml.jinja
```diff
@@ -48,7 +48,9 @@ services:
       - ./odoo/auto:/opt/odoo/auto:rw,z
     depends_on:
       - cdnjs_cloudflare_proxy
+      {% if postgres_version -%}
```
Yajo, line 51: Our code standard here for these kind of blocks is `{%- ... %}`.

Whitespace is weird in jinja and important in yaml.

Please follow it wherever possible (and here is possible). Same for all others.

Yajo (changes requested):
The patch is good, just one detail please.

devel.yaml.jinja
joao-p-marques: Ok. I actually did this following your suggestion in https://github.com/Tecnativa/doodba-copier-template/commit/abad70316b2bf8d924784615926c87fff6d0f1c2
I guess it doesn't apply the same here, as we are modifying a yaml array, and not adding the service, as before, right?

ska-ibees: Shouldn't we put the same condition on 'pgweb' dependency and on the 'db' under volumes as well?

Yajo: >  I guess it doesn't apply the same here, as we are modifying a yaml array, and not adding the service, as before, right?

Well, your code is not wrong here, but it was wrong there. :yum: 

the `-` is telling jinja to remove all whitespace from (or to) that boundary. Wherever possible, the code standard here is to remove previous whitespace, and here it is possible. In the other code you link, it was not possible (it removed one extra line).



> Shouldn't we put the same condition on 'pgweb' dependency and on the 'db' under volumes as well?

Yes, on every service that depends on `db` actually

prod.yaml.jinja
```diff
@@ -20,7 +20,9 @@ services:
       SMTP_SERVER: smtplocal
       {%- endif %}
     depends_on:
+      {% if postgres_version -%}
```
Yajo, line 23: Apply here the suggestion for https://github.com/Tecnativa/doodba-copier-template/pull/159#discussion_r513271120, sorry.

```suggestion
      {%- if postgres_version %}
```

test.yaml.jinja
```diff
@@ -20,7 +20,9 @@ services:
       SMTP_SERVER: smtplocal
     restart: unless-stopped
     depends_on:
+      {% if postgres_version -%}
```
Yajo, line 23: Test environment is supposed to always have a local postgres.

devel.yaml.jinja
Yajo, line 51: Devel environment is supposed to always have a local db.

Yajo (changes requested):

devel.yaml.jinja
ska-ibees: In that case, IMOH below choice should be removed from copier question:

`I will use an external PostgreSQL server: null`

And also it is not clear in docs that how to change DB parameters when deploying in production. It will be good to add some help for that.

joao-p-marques: > Nope. Only on prod.yaml.

In that case, if the  db service should always exist in devel and test, maybe we should end up with a different copier answer or mechanism to set up the case when we want an external DB on prod. If we set this var to `null`, we can't know which version to use in devel and test, right?

Maybe add another question with "Whether or not ti use external server?", that would override the db configuration only in prod.

ska-ibees: > Maybe add another question with "Whether or not ti use external server?", that would override the db configuration only in prod.
Or maybe split db questions for prod and dev/test. For instance, there are separate questions for production (domains_prod) and test domains (domains_test).

Yajo: Yes, we have to rethink the UX with this question.

joao-p-marques:
@Yajo I'm trying to ressurect this old PR with a different approach to the UX, following the suggestions above. WDYT?

Yajo:
Possibly all will be easier with copier 6 and `when` clause. IMHO we can wait until then.

PR #160, opened 2020-10-28
IMP: Add Odoo tests and install task

joao-p-marques:
1. Add invoke task to easily run Odoo tests. It accepts both the name of the module(s) to test, or it can detect the module you are currently working on if you pass a working directory or file. It also allows to run the tests in debug mode.
2. Add VSCode tasks that match invoke task and respective options. A normal task automatically passes to invoke the file you are working on.
3. Add VSCode debug configurations for debugging Odoo tests.
4. Add invoke task to easily install Odoo modules. You can pass the name(s), or also `--core`, `--extra` or `--private` options to install everyting.

It would also be good to incorporate https://github.com/Tecnativa/doodba-copier-template/pull/139 here, once it is finished.

TT26483

### DIFF
--- a/tasks.py.jinja
+++ b/tasks.py.jinja
@@ -17,6 +17,7 @@ from pathlib import Path
 from shutil import which
 
 from invoke import task
+from invoke import exceptions
 from invoke.util import yaml
 
 ODOO_VERSION = {{ odoo_version }}
@@ -357,10 +358,18 @@ def start(c, detach=True, debugpy=False):
     develop,
     help={
         "modules": "Comma-separated list of modules to test. Default: 'base'.",
+        "core": "Install all core addons. Default: False",
+        "extra": "Install all extra addons. Default: False",
+        "private": "Install all private addons. Default: False",
     },
 )
 def install(c, modules=None, core=False, extra=False, private=False):
     """Install Odoo addons"""
+    if not (modules or core or extra or private):
+        raise exceptions.ParseError(
+            message="You must provide at least one option. "
+            "See --help for details."
+        )
     cmd = "docker-compose run --rm addons init"
     if core:
         cmd += " --core"
@@ -379,40 +388,20 @@ def install(c, modules=None, core=False, extra=False, private=False):
         )
 
 
-def _get_cwd_addon(file=None):
-    if not file:
-        cwd = Path.cwd()
-    else:
-        cwd = Path.parent(file)
-    manifest_file = False
-    while not manifest_file:
-        manifest_file = (
-            Path(cwd / "__manifest__.py").exists()
-            or Path(cwd / "__openerp__.py").exists()
-        )
-        if manifest_file:
-            return cwd.stem
-        cwd = cwd.parent
-        if cwd == PROJECT_ROOT:
-            return None
-
-
 @task(
     develop,
     help={
-        "debugpy": "Whether or not to run tests in a VSCode debugging session. "
-        "Default: False",
+        "debugpy": "Whether or not to run tests in a VSCode debugging session. " \
+            "Default: False",
         "modules": "Comma-separated list of modules to test. Default: 'base'.",
-        "logs": "Whether or not to follow Odoo logs during testing. " "Default: True",
+        "logs": "Whether or not to follow Odoo logs during testing. " \
+            "Default: True",
     },
 )
-def test(c, modules=None, debugpy=False, logs=True, cur_file=None):
+def test(c, modules="base", debugpy=False, logs=True):
     """Run Odoo tests
     NOTE: Odoo must be restarted manually after this to go back to normal mode
     """
-    if not modules:
-        cur_module = _get_cwd_addon(file=cur_file)
-        modules = cur_module if cur_module is not None else "base"
     tmp_docker_compose_file = tempfile.NamedTemporaryFile(
         dir=PROJECT_ROOT, mode="w", suffix=".yaml"
     )
@@ -437,7 +426,8 @@ def test(c, modules=None, debugpy=False, logs=True, cur_file=None):
                     UID_ENV,
                 ),
             )
-            start(c) # If without logs, Odoo must be restarted manually
+    _logger.info("Waiting for services to spin up...")
+    time.sleep(SERVICES_WAIT_TIME)
     tmp_docker_compose_file.close()
 
 
--- a/tasks.py.jinja
+++ b/tasks.py.jinja
@@ -11,11 +11,13 @@ import json
 import os
 import time
 from logging import getLogger
+import tempfile
 from itertools import chain
 from pathlib import Path
 from shutil import which
 
 from invoke import task
+from invoke.util import yaml
 
 ODOO_VERSION = {{ odoo_version }}
 PROJECT_ROOT = Path(__file__).parent.absolute()
@@ -101,6 +103,12 @@ def write_code_workspace_file(c, cw_path=None):
                 "preLaunchTask": "Start Odoo in debug mode",
             },
             {
+                "name": "Run and debug Odoo tests",
+                "configurations": ["Attach Python debugger to running container"],
+                "preLaunchTask": "Run Odoo Tests in debug mode",
+                "internalConsoleOptions": "openOnSessionStart",
+            },
+            {
                 "name": "Start Odoo and debug JS in Firefox",
                 "configurations": ["Connect to firefox debugger"],
                 "preLaunchTask": "Start Odoo",
@@ -181,6 +189,42 @@ def write_code_workspace_file(c, cw_path=None):
                     "color": "#065535"
                     }
                 },
+            },
+            {
+                "label": "Run Odoo Tests for current module",
+                "type": "process",
+                "command": "invoke",
+                "args": ["test", "--cur-file", "${file}"],
+                "presentation": {
+                    "echo": True,
+                    "reveal": "always",
+                    "focus": True,
+                    "panel": "shared",
+                    "showReuseMessage": True,
+                    "clear": False,
+                },
+                "problemMatcher": [],
+            },
+            {
+                "label": "Run Odoo Tests in debug mode for current module",
+                "type": "process",
+                "command": "invoke",
+                "args": [
+                    "test",
+                    "--cur-file",
+                    "${file}",
+                    "--debugpy",
+                    "--no-logs",
+                ],
+                "presentation": {
+                    "echo": True,
+                    "reveal": "silent",
+                    "focus": False,
+                    "panel": "shared",
+                    "showReuseMessage": True,
+                    "clear": False,
+                },
+                "problemMatcher": [],
             },
             {
                 "label": "Start Odoo in debug mode",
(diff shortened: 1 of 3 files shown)


tasks.py.jinja
```diff
@@ -95,6 +97,12 @@ def write_code_workspace_file(c, cw_path=None):
                 "configurations": ["Attach Python debugger to running container"],
                 "preLaunchTask": "Start Odoo in debug mode",
             },
+            {
+                "name": "Run and debug Odoo tests",
+                "configurations": ["Attach Python debugger to running container"],
+                "preLaunchTask": "Run Odoo Tests in debug mode",
+                "internalConsoleOptions": "openOnSessionStart",
```
joao-p-marques, line 104: I wanted to include a `postDebugTask` here to automatically `Start Odoo` (restart in normal mode without debug) when you finish the tests, but VSCode `compounds` doesn't support it, and when I added it in the `Attach Python debugger to running container` configuration, I got an error saying I couldn't call a task from a different workspace.
I ended up leaving this like so, but it would be nice to have that feature

pedrobaeza: Why starting Odoo? If you launch this task, you don't want to end up with an started Odoo. You just want to run tests.

tasks.py.jinja
```diff
@@ -292,6 +344,59 @@ def start(c, detach=True, debugpy=False):
...
+                DOODBA_DEBUGPY_ENABLE=str(int(debugpy)),
+            ),
+        )
+        if logs:
+            c.run(
+                "docker-compose logs -f odoo",
+                env=dict(
+                    UID_ENV,
+                ),
+            )
```
joao-p-marques, line 383: I ended up with this approach because, if we run `docker-compose` in detached mode, we don't get the test logs and we don't know when the command ends.
If we run it it normal mode, we get the logs from all the services, and still don't know when the tests finished. 
This allows us to get both.

tasks.py.jinja
joao-p-marques: Ok.
It's just that, if you are used to running the tests with the docker-compose command, you are normally always running an Odoo instance, and then run another for the tests. So, when the tests finish, you end up with a started Odoo.
This way, if you had a running Odoo instance, this task needs to stop it and replace it with a new one for the tests, so you end up with a stopped Odoo. But I guess that is not a problem anyway...

pedrobaeza: No, I don't start Odoo "normally" when running tests. Or maybe yes, but I think it doesn't matter. This should be independent from that. Or there are technical limitations to this?

joao-p-marques: It is difficult to make it independent form that, because we need to map the debugger server port to the `odoo-proxy`, and when you run a container separately (with `docker-compose run --rm odoo --test-enable [...]`) it causes issues with the docker private network and the debugger port is not mapped, so we need to run it in the same Odoo container... That's why it can't be independent from what you had running.

pedrobaeza: OK, but there's a problem with that: when you run Odoo, you do it with several workers, and tests are not fine with workers.

joao-p-marques: Yes, when we run the tests in this case, we are relaunching the container replacing the Odoo start command, so we wouldn't have that problem...
Is that what you meant?

pedrobaeza: I suppose that

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+        )
+
+
+def _get_cwd_addon(file=None):
+    if not file:
+        cwd = Path.cwd()
+    else:
+        cwd = Path.parent(file)
+    manifest_file = False
+    while not manifest_file:
```
Yajo, line 397: This check is not used, because if a `manifest_file` is found, the `return` exits the `while`.

Instead:

```suggestion
    while PROJECT_ROOT < cwd:
```

And remove the same test you do at the end of the loop.

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+        manifest_file = (
+            Path(cwd / "__manifest__.py").exists()
+            or Path(cwd / "__openerp__.py").exists()
```
Yajo, line 400:
```suggestion
            (cwd / "__manifest__.py").exists()
            or (cwd / "__openerp__.py").exists()
```

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+                UID_ENV,
+            ),
+        )
+
+
+def _get_cwd_addon(file=None):
+    if not file:
+        cwd = Path.cwd()
+    else:
+        cwd = Path.parent(file)
```
Yajo, line 395: According to your loop below, this is the same but simpler:

```suggestion
def _get_cwd_addon(file="."):
    cwd = Path(file)
```

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+@task(
+    develop,
+    help={
+        "debugpy": "Whether or not to run tests in a VSCode debugging session. "
+        "Default: False",
+        "modules": "Comma-separated list of modules to test. Default: 'base'.",
+        "logs": "Whether or not to follow Odoo logs during testing. " "Default: True",
+    },
+)
+def test(c, modules=None, debugpy=False, logs=True, cur_file=None):
```
Yajo, line 418: Missing mode. Testing on init or on update differs. Default to init.

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+    """Run Odoo tests
+    NOTE: Odoo must be restarted manually after this to go back to normal mode
+    """
+    if not modules:
+        cur_module = _get_cwd_addon(file=cur_file)
+        modules = cur_module if cur_module is not None else "base"
```
Yajo, line 424: You don't need to be so specific in Python. Just:

```py
        modules = cur_module or "base"
```

However, as said above, I think "base" is not a good default.

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+
+def _override_docker_command(c, service, command, file=None):
+    docker_config = {
+        "version": "2.4",  # TODO: Update this if main file is updated
+        "services": {service: {"command": command.split(" ")}},
+    }
+    docker_config_yaml = yaml.dump(docker_config)
+    if file:
+        file.write(docker_config_yaml)
+        file.flush()
```
Yajo, line 462: Is this function useful for something without `file`? :thinking:

Yajo (changes requested):

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+        if cwd == PROJECT_ROOT:
+            return None
+
+
+@task(
+    develop,
+    help={
+        "debugpy": "Whether or not to run tests in a VSCode debugging session. "
+        "Default: False",
+        "modules": "Comma-separated list of modules to test. Default: 'base'.",
```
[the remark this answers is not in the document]
joao-p-marques: Ok, I agree. I ended up by running the tests against the module you are currently working on (based on the CWD) and, if none are found, the task will fail. WDYT?

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+                ),
+            )
+    _logger.info("Waiting for services to spin up...")
+    time.sleep(SERVICES_WAIT_TIME)
+    tmp_docker_compose_file.close()
+
+
+def _override_docker_command(c, service, command, file=None):
+    docker_config = {
+        "version": "2.4",  # TODO: Update this if main file is updated
```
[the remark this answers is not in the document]
joao-p-marques: Yes, done. Thanks for the suggestion

tasks.py.jinja
joao-p-marques: Actually, I need the **parent** if you pass a file, and the cwd if not. So I think it is better to do this verification. OTOH, I simplified the line to:
```
cwd = Path(file).parent if file else Path.cwd()
```

tasks.py.jinja
joao-p-marques: No, indeed. I was thinking of other uses when I wrote it but it ended up only being useful with an output file. Fixed. TY!

tasks.py.jinja
```diff
@@ -11,21 +11,64 @@ import json
...
+SERVICES_WAIT_TIME = int(os.environ.get("SERVICES_WAIT_TIME", 4))
 
 _logger = getLogger(__name__)
 
 
+def _override_docker_command(service, command, file, orig_file=None):
+    # Read config from main file
+    if orig_file:
+        with open(orig_file, "r") as fd:
+            orig_docker_config = yaml.load(fd.read())
```
Yajo, line 36:
```suggestion
            orig_docker_config = yaml.safe_load(fd.read())
```

tasks.py.jinja
```diff
@@ -11,21 +11,64 @@ import json
...
+        "services": {service: {"command": command}},
+    }
+    docker_config_yaml = yaml.dump(docker_config)
+    file.write(docker_config_yaml)
+    file.flush()
+
+
+def _remove_auto_reload(file, orig_file):
+    with open(orig_file, "r") as fd:
+        orig_docker_config = yaml.load(fd.read())
```
Yajo, line 51:
```suggestion
        orig_docker_config = yaml.safe_load(fd.read())
```

tasks.py.jinja
```diff
@@ -11,21 +11,64 @@ import json
...
+    odoo_command = orig_docker_config["services"]["odoo"]["command"]
+    odoo_command[-1] = odoo_command[-1].replace("reload,", "")
```
Yajo, line 53: You can't do this so blindly. Better search the element that starts with `"--dev"` and modify it.

tasks.py.jinja
```diff
@@ -302,13 +388,135 @@ def lint(c, verbose=False):
...
+            mode="w",
+            suffix=".yaml",
+        )
+        cmd = (
+            "docker-compose -f docker-compose.yml "
+            f"-f {tmp_docker_compose_file.name} up"
+        )
+        _remove_auto_reload(
+            tmp_docker_compose_file,
+            orig_file=Path(str(PROJECT_ROOT), "docker-compose.yml"),
```
Yajo, line 403:
```suggestion
            orig_file=PROJECT_ROOT / "docker-compose.yml",
```

tasks.py.jinja
```diff
@@ -302,13 +388,135 @@ def lint(c, verbose=False):
...
+            modules = cur_module
+    tmp_docker_compose_file = tempfile.NamedTemporaryFile(
+        mode="w",
+        suffix=".yaml",
+    )
+    cmd = (
+        "docker-compose -f docker-compose.yml "
+        f"-f {tmp_docker_compose_file.name} up -d"
+    )
+    odoo_command = ["odoo", "--test-enable"]
```
Yajo, line 491:
```suggestion
    odoo_command = ["odoo", "--test-enable", "--stop-after-init", "--workers=0"]
```

tasks.py.jinja
```diff
@@ -302,13 +388,135 @@ def lint(c, verbose=False):
...
+    if mode == "init":
+        odoo_command.append("-i")
+    elif mode == "update":
+        odoo_command.append("-u")
+    else:
+        raise exceptions.ParseError(
+            message="Available modes are 'init' or 'update'. See --help for details."
+        )
+    odoo_command.extend([modules, "--stop-after-init", "--workers=0"])
```
Yajo, line 500:
```suggestion
```

tasks.py.jinja
```diff
@@ -309,6 +354,115 @@ def start(c, detach=True, debugpy=False):
...
+        if logs:
+            c.run(
+                "docker-compose logs -f odoo",
+                env=dict(
+                    UID_ENV,
+                ),
+            )
+    _logger.info("Waiting for services to spin up...")
+    time.sleep(SERVICES_WAIT_TIME)
+    tmp_docker_compose_file.close()
```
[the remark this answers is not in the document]
Yajo: This remains unattended.

Yajo (changes requested):

tests/test_tasks_downstream.py
```diff
@@ -128,6 +139,116 @@ def test_start(
...
+            # Check if auto-reload is disabled
+            container_logs = docker_compose("logs", "odoo")
+            assert "dev=reload" not in container_logs
+    finally:
+        # Imagine the user is in the odoo subrepo for this command
+        with local.cwd(tmp_path / "odoo" / "custom" / "src" / "odoo"):
+            invoke("stop", "--purge")
+
+
+@pytest.mark.sequential
```
Yajo, line 151: Is this test really sequential? It doesn't seem so... :thinking: 

```suggestion
```

Yajo (changes requested):

tests/test_tasks_downstream.py
joao-p-marques: Well, it requires bringing up the Odoo container, which depends on all the others that were causing problems with the `start` task tests.

Yajo: Yes but it opens no ports.

Yajo (approved):

PR #161, opened 2020-10-29
Add buttons to run start and stop tasks

joao-p-marques:
Add [Tasks extension](https://marketplace.visualstudio.com/items?itemName=actboy168.tasks) to allow launching Start and Stop Odoo tasks from the bottom tasks bar. 

I've added it with some icons and color, but we can also leave them blank and add some text to describe.


joao-p-marques:
This is how it would look:

image

Yajo (approved):

PR #168, opened 2020-11-09
Use a real python file for downstream tasks

Yajo:
This will make `black` and other linters help us.

The only fix needed was to obtain `ODOO_VERSION` from `common.yaml` instead of directly from the template value.

This is the strategy followed before 9855d7ad129e2c8e82bbb588574d15e16071893e and it was working fine, so let's return to it.


tasks_downstream.py
```diff
@@ -0,0 +1,580 @@
...
+                "localRoot": "${workspaceRoot:%s}" % subrepo.name,
+                "remoteRoot": f"/opt/odoo/custom/src/{subrepo.name}",
+            }
+        )
+        for addon in chain(subrepo.glob("*"), subrepo.glob("addons/*")):
+            if (addon / "__manifest__.py").is_file() or (
+                addon / "__openerp__.py"
+            ).is_file():
+                url = f"http://localhost:{ODOO_VERSION:.0f}069/{addon.name}/static/"
+                path = "${{ '{{' }}workspaceRoot:{}{{ '}}' }}/{}/static/".format(
```
joao-p-marques, line 206: Is this something left from the jinja template?
```suggestion
                path = "${{workspaceRoot:{}}}/{}/static/".format(
```

joao-p-marques (commented):
This looks perfect. I will make it much easier to do fixes on the template.

I just have a small comment. Don't know if it is an error or if I'm just seeing it wrong :sweat_smile:

tasks_downstream.py
```diff
@@ -0,0 +1,580 @@
...
+            cw_config = json.load(cw_fd)
+    except (FileNotFoundError, json.decoder.JSONDecodeError):
+        pass  # Nevermind, we start with a new config
+    # Launch configurations
+    debugpy_configuration = {
+        "name": "Attach Python debugger to running container",
+        "type": "python",
+        "request": "attach",
+        "pathMappings": [{"localRoot": f"{root_var}/odoo", "remoteRoot": "/opt/odoo"}],
+        "port": int(ODOO_VERSION) * 1000 + 899,
```
joao-p-marques, line 113: Also, seems like [tests are failing](https://github.com/Tecnativa/doodba-copier-template/pull/168/checks?check_run_id=1373860930#step:13:582) because of this conversion. This could be a workaround:
```suggestion
        "port": int(float(ODOO_VERSION)) * 1000 + 899,
```

Yajo: Thanks, solved in https://github.com/Tecnativa/doodba-copier-template/pull/168/files#diff-9fccd0ad3d50013f36bd06d3ccede328be22fe7d5d1009f445ff8d7f6de0f451R23 instead.

joao-p-marques (approved):

PR #170, opened 2020-11-10
Move search.followSymlinks to workspace setting

Yajo:
It turns out this setting is ignored if set as folder setting.


joao-p-marques (approved):

pedrobaeza (commented):
I don't know about the current structure, but I suppose it's correct

PR #171, opened 2020-11-10
Do not color task launchers

Yajo:
They were hard to see.

Before: 
imagen

After: 
imagen

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -228,9 +228,7 @@ def write_code_workspace_file(c, cw_path=None):
                     "clear": False,
                 },
                 "problemMatcher": [],
-                "options": {
-                    "statusbar": {"label": "$(play-circle)", "color": "#065535"}
-                },
+                "options": {"statusbar": {"label": "$(play-circle)"}},
             },
             {
                 "label": "Run Odoo Tests for current module",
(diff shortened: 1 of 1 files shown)


tasks_downstream.py
```diff
@@ -228,9 +228,7 @@ def write_code_workspace_file(c, cw_path=None):
                     "clear": False,
                 },
                 "problemMatcher": [],
-                "options": {
-                    "statusbar": {"label": "$(play-circle)", "color": "#065535"}
-                },
+                "options": {"statusbar": {"label": "$(play-circle)"}},
```
joao-p-marques, line 231:
```suggestion
                "options": {"statusbar": {"label": "$(play-circle) Start Odoo"}},
```

tasks_downstream.py
```diff
@@ -299,9 +297,7 @@ def write_code_workspace_file(c, cw_path=None):
                     "clear": False,
                 },
                 "problemMatcher": [],
-                "options": {
-                    "statusbar": {"label": "$(stop-circle)", "color": "#9A0000"}
-                },
+                "options": {"statusbar": {"label": "$(stop-circle)"}},
```
joao-p-marques, line 300:
```suggestion
                "options": {"statusbar": {"label": "$(stop-circle) Stop Odoo"}},
```

joao-p-marques (commented):
Yes, I was also "afraid" of that when I opened https://github.com/Tecnativa/doodba-copier-template/pull/161 :sweat_smile: 

I would suggest adding a small text on the side, which would make it more readable, and would also look better now that they are not coloured, but that is nitpicking :smile:

joao-p-marques (approved):

PR #172, opened 2020-11-10
Proxy www.googleapis.com

Yajo:
[It is needed when developing for `website_slides`][1].

I also refactored a bit how these proxies are generated. This renames them, but it shouldn't be a big problem because it is just in the devel environment. Now it'll be easier to extend.

[1]: https://github.com/odoo/odoo/blob/62a18bdb372e234107fd2099a2df5f9e3a442c96/addons/website_slides/models/slide_slide.py#L847


joao-p-marques (approved):
Looks good :ok_hand:

PR #176, opened 2020-11-11
Debugger fixes and improvements

joao-p-marques:
1. Map Python debugger to odoo/auto: All Odoo addons and core code are now mapped individually with the Python debugger.
2. Enable logs for test task when not in debug mode
3. Improve error messages

Error handling can still be improved, by trying to integrate better with VSCode tasks errors, but I had no success so far.

TT26728

### DIFF
--- a/tests/test_tasks_downstream.py
+++ b/tests/test_tasks_downstream.py
@@ -36,26 +36,20 @@ def _wait_for_test_to_start():
     return stdout
 
 
-def _tests_ran(stdout, odoo_version, addon_name, mode="-i"):
-    if (
-        f"Executing odoo --test-enable --stop-after-init --workers=0 {mode} {addon_name}"
-        not in stdout
-    ):
-        return False
-    if odoo_version > 13.0:
-        match_str = re.compile(r"devel\sodoo\.service\.server:\s.*\spost-tests\sin")
-        if not match_str.search(stdout):
-            return False
-    elif odoo_version > 10.0:
-        if "devel odoo.service.server: All post-tested" not in stdout:
-            return False
-    elif odoo_version > 9.0:
-        if "devel odoo.modules.loading: All post-tested" not in stdout:
-            return False
-    else:
-        if "devel openerp.modules.loading: All post-tested" not in stdout:
-            return False
-    return True
+def _tests_ran(stdout, odoo_version, addon_name):
+    # Ensure the addon was installed/updated, and not independent ones
+    assert f"module {addon_name}: creating or updating database tables" in stdout
+    # Ensure the addon was tested
+    main_pkg, suffix = "odoo", r"\:\sStarting"
+    if odoo_version < 13.0:
+        suffix = r"\srunning tests."
+    if odoo_version < 10.0:
+        main_pkg = "openerp"
+    assert re.search(fr"{main_pkg}\.addons\.{addon_name}\.tests\.\w+{suffix}", stdout)
+    # Check no alien addons are installed, updated or tested
+    if addon_name != "base":
+        assert "module base: creating or updating database tables" not in stdout
+        assert not re.search(fr"{main_pkg}\.addons\.base\.tests\.\w+{suffix}", stdout)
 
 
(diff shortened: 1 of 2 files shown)


Yajo (changes requested):

joao-p-marques:
Changes made @Yajo 

I had to test other addons to include other cases, but I tried to keep it simple and light, and test and install tasks are now tested together.

I also ended up ignoring the return code from the tests, so that our CI never fails because of some bad integration errors upstream, as has happened with https://github.com/Tecnativa/doodba-copier-template/runs/1389417193?check_suite_focus=true#step:14:988

tests/test_tasks_downstream.py
```diff
@@ -177,75 +177,48 @@ def test_install(
...
+                assert "Executing odoo --stop-after-init --init utm" in stdout
+                assert _install_status("mail") == "installed"
+                assert _install_status("utm") == "installed"
+                assert _install_status("note") == "uninstalled"
+            # Test "note" simple call in init mode (default)
+            stdout = invoke("test", "-m", "note", "--mode", "init", retcode=None)
+            assert (
+                "Executing odoo --test-enable --stop-after-init --workers=0 -i note"
+                in stdout
             )
```
Yajo, line 201: I think this'd be important:

```suggestion
            )
            assert "devel odoo.modules.loading: All post-tested" in stdout
```

tests/test_tasks_downstream.py
```diff
@@ -177,75 +177,48 @@ def test_install(
...
-            invoke("test", "-m", "purchase")
-            stdout = _wait_for_test_to_start()
+            assert _install_status("note") == "installed"
+            # Test "note" simple call in update mode
+            stdout = invoke("test", "-m", "note", "--mode", "update", retcode=None)
             assert (
-                "Executing odoo --test-enable --stop-after-init --workers=0 -i purchase"
+                "Executing odoo --test-enable --stop-after-init --workers=0 -u note"
                 in stdout
             )
```
Yajo, line 208: Here too:

```suggestion
            )
            assert "devel odoo.modules.loading: All post-tested" in stdout
```

Yajo (changes requested):
Almost there

Yajo:
Review the failure. Maybe the test is wrong or maybe odoo is really not testing the addons.

joao-p-marques:
Yes, I am already on it. Seems like the correct log is
```
devel odoo.service.server: All post-tested
```
OTOH, I am also getting different logs in v14. I am investigating if this logs change between versions...

Yajo:
> I am investigating if this logs change between versions...

Most probably yes, because I copied https://github.com/Tecnativa/doodba-copier-template/pull/176#discussion_r522024064 from v12.

joao-p-marques:
Yes, I ended up creating the `_tests_ran` function to check specific logs base on the current version. Let's see if it works for all the versions.

Yajo (approved):

tests/test_tasks_downstream.py
```diff
@@ -201,31 +195,28 @@ def test_install_test(
...
             # Test "note" simple call in init mode (default)
+            assert _install_status("note") == "uninstalled"
             stdout = invoke("test", "-m", "note", "--mode", "init", retcode=None)
             # Ensure "note" was installed and tests ran
             assert _install_status("note") == "installed"
             assert _tests_ran(stdout, supported_odoo_version, "note")
             # Test "note" simple call in update mode
             stdout = invoke("test", "-m", "note", "--mode", "update", retcode=None)
-            assert _tests_ran(stdout, supported_odoo_version, "note", mode="-u")
+            assert _tests_ran(stdout, supported_odoo_version, "note")
```
joao-p-marques, line 219: If you are doing `assert`s in the _tests_ran function, this needs to be adapted. Either return true or false in the function, or just do a normal call here.

joao-p-marques:
Just a small doubt... Otherwise tests look much better now, thank you!

PR #177, opened 2020-11-11
Fixes WSL2 support

theangryangel:
Binding to 127.0.0.1 makes the services unavailable to Windows when running Doodba under WSL(2).

This is a backport from our internal fork which has diverged slightly due to different needs. Apologies if I've missed something.

### DIFF
--- a/devel.yaml.jinja
+++ b/devel.yaml.jinja
@@ -10,8 +10,8 @@ services:
       default:
       public:
     ports:
-      - "127.0.0.1:{{ macros.version_major(odoo_version) }}899:6899"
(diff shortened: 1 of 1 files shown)


Yajo:
This is not a safe default. If you're working in an untrusted network, such as a café or code sprint, anybody could connect to your development environment.

Sorry, we cannot accept this. :man_shrugging:

joao-p-marques:
Yes, perhaps you can find a way to make the service available to Windows without exposing it like this? This seems to be a limitation of WSL anyway...

PR #178, opened 2020-11-16
Add restart options to reload environment

joao-p-marques:
After auto-reload as been removed, it is more difficult to debug when making regular changes. Start task does not recreate Odoo container, so no changes are reflected.

This ensures Odoo is recreated after every start command.
It also adds a VSCode task and button to easily restart Odoo and the proxy.

It would be nice to do it automatically when you press the "Restart" button in the VSCode debugger, but AFAICS that only restarts the "Attach" configurations, which will always find the debugger already running, leaving no option but to "Disconnect" and launch again. 
If you press "Restart Odoo", the debugging session is also disconnected and needs to be lauched again.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -315,6 +315,22 @@ def write_code_workspace_file(c, cw_path=None):
                 "problemMatcher": [],
                 "options": {"statusbar": {"label": "$(stop-circle) Stop Odoo"}},
             },
+            {
+                "label": "Restart Odoo",
+                "type": "process",
+                "command": "invoke",
+                "args": ["restart"],
+                "presentation": {
+                    "echo": True,
+                    "reveal": "silent",
+                    "focus": False,
+                    "panel": "shared",
+                    "showReuseMessage": True,
+                    "clear": False,
+                },
+                "problemMatcher": [],
+                "options": {"statusbar": {"label": "$(history) Restart Odoo"}},
+            },
         ],
     }
     # Sort project folders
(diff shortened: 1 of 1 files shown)


tasks_downstream.py
```diff
@@ -416,7 +432,16 @@ def start(c, detach=True, debugpy=False):
...
-            c.run(cmd, env=dict(UID_ENV, DOODBA_DEBUGPY_ENABLE=str(int(debugpy))))
+            result = c.run(
+                cmd,
+                pty=True,
+                env=dict(
+                    UID_ENV,
+                    DOODBA_DEBUGPY_ENABLE=str(int(debugpy)),
+                ),
+            )
+            if not ("Recreating" or "Starting" in result.stdout):
```
Yajo, line 443: This won't work as you expect:

```py
>>> "Recreating" or "Starting" in "hello"
'Recreating'
>>> bool("Recreating" or "Starting" in "hello")
True
```

Yajo (changes requested):

tasks_downstream.py
joao-p-marques: Oh of course :man_facepalming: 
Sorry, MB. Changes made!

joao-p-marques:
@Yajo tests are failing due to some other problem.
Seems like the `set-env` option in GH Actions is not deprecated anymore, it is disabled :sweat_smile:

Yajo (approved):

PR #185, opened 2020-11-24
Add option to skip uninstallable addons from pre-commit hooks

joao-p-marques:
Apply same as in https://github.com/OCA/oca-addons-repo-template/pull/3

Usage example: https://github.com/OCA/mis-builder/commit/38f75413687e79867840c9e6f86ee207394fed86#diff-82c3b36709926de3a1b27bea0d8258cc

@Tecnativa
TT26700


Yajo:
I set to draft until this has tests for modules inside private.

joao-p-marques:
Tests added. Just to clarify: this will only apply to **private** addons.

.pre-commit-config.yaml.jinja
```diff
@@ -68,6 +70,14 @@ repos:
       - id: check-xml
       - id: mixed-line-ending
         args: ["--fix=lf"]
+  - repo: https://github.com/oca/maintainer-tools
+    rev: 2c66e72 # Leave as is for stability
+    hooks:
+      # update the NOT INSTALLABLE ADDONS section above
```
Yajo, line 76: I think this hook should be the 1st or 2nd one, to make all other hooks skip this code.

tests/test_nitpicking.py
```diff
@@ -427,6 +427,110 @@ def some_method(self, test):
...
+                    class ResPartner(models.Model):
+                        _name = "res.partner"
+
+                        def some_method(self, test):
+                            """some weird
+                            docstring"""
+                            _logger.info(models, join, get, io, sys, odoo)
+                ''',
+            }
+        )
```
Yajo, line 473: You might consider moving this to a helper in this module, as it is used several times.

tests/test_nitpicking.py
```diff
@@ -427,6 +427,110 @@ def some_method(self, test):
...
+        git("add", "-A")
+        # First commit will add new module to the exclude list in pre-commit
+        git("commit", "-m", "added test_module_2", retcode=1)
+        git("commit", "-am", "added test_module_2")
```
Yajo, line 477: I think you have to check that there are no changes inside the module that cannot be installed.

joao-p-marques:
Ty, changes made.

.pre-commit-config.yaml.jinja
```diff
@@ -20,6 +22,14 @@ repos:
         entry: found forbidden files; remove them
         language: fail
         files: "\\.rej$"
+  - repo: https://github.com/oca/maintainer-tools
+    rev: f1b5595 # Leave as is for stability
```
Yajo, line 26: That comment is a bit stupid. If it were for that, we'd leave as it is all hooks :shrug: 

```suggestion
    rev: f1b5595
```

joao-p-marques: Well, what I meant was that, in this particular hook, we don't have any branches or tags to refer to, and would have to always pick from master. By referring to a particular commit, we make sure we don't get unexpected changes, thus "increasing stability".
In all the other hooks we don't need to because we refer to tagged versions.

I have removed it anyway :smile:

tests/test_nitpicking.py
```diff
@@ -427,6 +402,39 @@ def some_method(self, test):
...
+        # First commit will add new module to the exclude list in pre-commit
+        git("commit", "-m", "start migration of test_module", retcode=1)
+        # Module should now be ignored by pre-commit and give no problems in commit
+        git("commit", "-am", "start migration of test_module")
+        # Load pre-commit config
+        with open(tmp_path / ".pre-commit-config.yaml", "r") as fd:
+            pre_commit_config = yaml.safe_load(fd.read())
+        assert "^odoo/custom/src/private/test_module/|" in pre_commit_config["exclude"]
+        # Make sure uninstallable addon was ignored by pre-commit
+        assert "test_module" not in git("status", "--porcelain")
```
Yajo, line 421: I think this is not the test you want. There are no changes in that folder since last commit, so obviously git status will report nothing there.

What you really mean is to check that `pre-commit run -a` passes. Remember that, before doing a commit, what git runs is `pre-commit run` without the `-a`.

tests/conftest.py
```diff
@@ -183,3 +183,85 @@ def socket_is_open(host, port):
...
+                class ResPartner(models.Model):
+                    _name = "res.partner"
+
+                    def some_method(self, test):
+                        """some weird
+                        docstring"""
+                        _logger.info(models, join, get, io, sys, odoo)
+            ''',
+        }
+    )
```
Yajo, line 267: I think that these helpers are very useful. However I think that they are very related among themselves so as to be a single function `generate_test_addon(..., ugly=True/False)`.

Yajo (changes requested):

tests/conftest.py
Yajo: Add a little docstring too please.

joao-p-marques:
Changes made @Yajo

tests/conftest.py
```diff
@@ -183,3 +183,84 @@ def socket_is_open(host, port):
...
+            from logging import getLogger
+            import io,sys,odoo
+            _logger=getLogger(__name__)
+            class ResPartner(models.Model):
+                _name='res.partner'
+                def some_method(self,test):
+                    '''some weird
+                        docstring'''
+                    _logger.info(models,join,get,io,sys,odoo)
+        """
```
Yajo, line 265: This code is inverted. However, let me push the change.

Yajo (approved):

tests/conftest.py
joao-p-marques: Oh sorry :man_facepalming:

PR #189, opened 2020-11-26
Fix regex from Traefik 1 path prefix rules

### LINKED ISSUE #166: Traefik 1 labels seem to prevent Website editing
So I was getting all sorts of errors trying to click edit on the frontend website in prod but not dev. Tried a few things, including a copier update -D.

Eventually just deleted all the Traefik labels I thought could cause it as well as all the Traefik 2 labels and it worked. Then I noticed there was 301 redirect happening for a POST to `/web/dataset/call_kw` to load the Wysiwyg Editor assets. Which was creating a typical endpotini error about calling with http instead of JSON.

Working back through the labels. Eventually we found this one causes the issue, and commenting out it works, although actually I think it might not be the real culprit, although I don't know. For avoidance of doubt <your host> is anonymisation not the actual label.

```yaml
traefik.forbiddenCrawlers-0.frontend.rule: Host:<your host>;Path:/web{anything:/.*},/website/info{anything:/.*}
```

EDIT: Sorry missing details Traefik 1.7, 13.0 Doodba. Pretty Vanilla setup in terms of extra mods and the like, all recently updated.

Yajo:
The wise man once said:

> You have a problem. You use a regexp to fix it. Now you have two problems.

Let's ~~avoid~~ fix regexes to fix #166.

BTW I update MQT repo.

### DIFF
--- a/_traefik1_labels.yml.jinja
+++ b/_traefik1_labels.yml.jinja
@@ -10,8 +10,11 @@
 {%- macro path_prefix_rule(path_prefixes) -%}
   Path:
     {%- for path in path_prefixes %}
-      {%- set sep = "" if path.endswith("/") else "/" %}
-      {{- path }}{anything:{{ sep }}.*}
+      {%- if path.endswith("/") %}
+        {{- path }}{anything:.*}
+      {%- else %}
+        {{- path }},{{ path }}/{anything:.*}
+      {%- endif %}
       {%- if not loop.last %},{% endif %}
     {%- endfor %}
 {%- endmacro %}
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -24,7 +24,9 @@ def test_multiple_domains(
     # XXX Remove traefik1 specific stuff some day
     is_traefik1 = version.parse(traefik_host["traefik_version"]) < version.parse("2")
     data = {
+        "odoo_listdb": True,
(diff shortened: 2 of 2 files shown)


Yajo:
Oops, it turns out [the regexp is required](url). Otherwise we need to add several rules to match i.e. both `Path:/web` and `PathPrefix:/web/`. [This test error](https://github.com/Tecnativa/doodba-copier-template/pull/189/checks?check_run_id=1458680835#step:13:124) shows it:

```
>                   assert response.headers["X-Robots-Tag"] == "noindex, nofollow"

tests/test_routing.py:87: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = {'Content-Encoding': 'gzip', 'Content-Type': 'text/html; charset=utf-8', 'Date': 'Thu, 26 Nov 2020 11:56:49 GMT', 'Ser...MT; Max-Age=7776000; HttpOnly; Path=/', 'Vary': 'Accept-Encoding', 'X-Frame-Options': 'DENY', 'Content-Length': '1155'}
key = 'X-Robots-Tag'

    def __getitem__(self, key):
>       return self._store[key.lower()][1]
E       KeyError: 'x-robots-tag'
```

PR #192, opened 2020-12-04
Add Python auto-complete configuration

joao-p-marques:
Provide the Odoo path to the auto complete engine in VSCode.

Also default to Jedi as Python Language Server as, for now, it is the only one that supports auto-complete with Odoo

TT27154


tasks_downstream.py
```diff
@@ -108,7 +108,33 @@ def write_code_workspace_file(c, cw_path=None):
...
-    cw_config["settings"].update({"search.followSymlinks": False})
+    cw_config["settings"].update(
+        {
+            "editor.formatOnSave": True,
+            "python.autoComplete.extraPaths": [f"{str(SRC_PATH)}/odoo"],
+            "python.formatting.blackArgs": [],
+            "python.formatting.provider": "black",
+            "python.languageServer": "Jedi",
+            "python.linting.flake8Enabled": True,
+            "python.linting.ignorePatterns": [f"{str(SRC_PATH)}/odoo/**/*.py"],
```
joao-p-marques, line 119: Just to clarify: I tried to use the `${workspaceFolder}` variable here but, for some reason, it only works with this full path :man_shrugging:

tasks_downstream.py
```diff
@@ -108,7 +108,33 @@ def write_code_workspace_file(c, cw_path=None):
         pass  # Nevermind, we start with a new config
     # Static settings
     cw_config.setdefault("settings", {})
-    cw_config["settings"].update({"search.followSymlinks": False})
+    cw_config["settings"].update(
+        {
+            "editor.formatOnSave": True,
```
Yajo, line 113: Beware. Enabling this in the `odoo` folder will break your workflow when pushing a PR to odoo.

Also, we don't want to `black` OCA modules in versions prior to 13.0.

Please check those 2 points and report back.

joao-p-marques: Yes, I am working on that :smiley: Thanks!

joao-p-marques:
@Yajo I ended up removing the automatic formatting bellow v13 because it would be impossible to filter the files that we want to be formatted and the ones that we don't. 
As oppose to linters, in VSCode, you can't specify a path to be ignored by the formatter, at least until https://github.com/microsoft/vscode-python/issues/8639 is solved.
OTOH, the way `black` resolves file paths brings some constraints. As of now, VSCode passes the full file path, and black computes a relative path to the first folder in the chain that contains a `.git` folder or `pyproject.toml`, so we can only filter by files inside the `odoo` or `addons` folders, which should always be in the odoo project itself, and not on OCA repos.
WDYT?

Yajo:
> we can only filter by files inside the `odoo` or `addons` folders, which should always be in the odoo project itself, and not on OCA repos.

That seems like a reasonable workaround.

Yajo:
I'm thinking that another approach would be to just disable `formatOnSave`. Anyways, pre-commit will do it. I understand it's more comfortable to have it enabled, but if it's going to introduce too many headaches... :man_shrugging:

joao-p-marques:
Yes, that is true and a good option. I was just thinking it is, indeed, a
nice feature 😁, but also that someone might enable it globally and end up
reformatting all the files, so this solution would stop that. Anyway, it is
not important TBH, and maybe just leaving it as is is the best option.

Yajo:
So is this your final code?

joao-p-marques:
Sorry, I was making the changes :smile: Now yes

pedrobaeza:
Personally I use format on save for having the code in their final form. Why is that problematic?

joao-p-marques:
It ends up reformatting Odoo code and causing massive diffs... :man_shrugging: And the same in OCA bellow v13

PR #201, opened 2020-12-14
Add migration script with tests to move from settings.json to workspace file

joao-p-marques:
TT27154


joao-p-marques:
I think this is ready @Yajo 
Tested locally and everything seems to work. We would just need to release `v2.7.0a4` for this.

Yajo (approved):

PR #217, opened 2021-02-12
Migrate images to ghcr + Small improvements

joao-p-marques:
Images migrated to GHCR so far:
- docker-duplicity 
- docker-mailserver (the image we were using had been deprecated and might have been outdated. I referenced the stable release, but we should still do some testing before releasing it officially)
- docker-whitelist (depends on https://github.com/Tecnativa/docker-whitelist/pull/6)

Still waiting on response from other maintainers to have other images migrated as well, but this should reduce the impact of the restrictions of DockerHub already.

Also, some small improvements (suggestions):

1. Add a button to see container logs:
image
As happens with the `Test module` button, it launches the task and automatically focuses the terminal there to see the logs.
2. The `invoke start` command was recreating the containers even if they had just been created. This prevents that.
3. Add 2 simple badges at the README of the generated project pointing to the `prod` and `test` deployment environments, if defined. 
image

ping @Yajo

### DIFF
--- a/README.md.jinja
+++ b/README.md.jinja
@@ -1,9 +1,16 @@
+{%- import "_macros.jinja" as macros -%}
 [![Doodba deployment](https://img.shields.io/badge/deployment-doodba-informational)](https://github.com/Tecnativa/doodba)
 [![Last template update](https://img.shields.io/badge/last%20template%20update-{{ _copier_answers._commit|replace('-', '--') }}-informational)](https://github.com/Tecnativa/doodba-copier-template/tree/{{ _copier_answers._commit }})
 [![Odoo](https://img.shields.io/badge/odoo-v{{ odoo_version }}-a3478a)](https://github.com/odoo/odoo/tree/{{ odoo_version }})
 {%- if gitlab_url %}
 [![pipeline status]({{ gitlab_url }}/badges/{{ odoo_version }}/pipeline.svg)]({{ gitlab_url }}/commits/{{ odoo_version }})
 [![coverage report]({{ gitlab_url }}/badges/{{ odoo_version }}/coverage.svg)]({{ gitlab_url }}/commits/{{ odoo_version }})
+{%- endif %}
+{%- if domains_prod %}
+[![Deployment data](https://img.shields.io/badge/Deployment-prod-green)](http://{{ macros.first_main_domain(domains_prod)|tojson }})
+{%- endif %}
+{%- if domains_test %}
+[![Deployment data](https://img.shields.io/badge/Deployment-demo-yellow)](http://{{ macros.first_main_domain(domains_test)|tojson }})
 {%- endif %}
 {%- if project_license %}
 [![{{ project_license }} license](https://img.shields.io/badge/license-{{ project_license|replace("-", "--") }}-{%- if project_license in ["OEEL-1.0", "OPL-1.0"] %}critical{%- else %}success{%- endif %}})](LICENSE)
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -50,7 +50,7 @@ services:
   {%- if smtp_relay_host %}
 
   smtpreal:
-    image: tvial/docker-mailserver:latest
+    image: ghcr.io/docker-mailserver/docker-mailserver:{{ smtp_relay_version }}
     domainname: "{{ smtp_canonical_default }}"
     stop_grace_period: 1m
     volumes:
@@ -83,7 +83,7 @@ services:
   {%- if backup_dst %}
 
   backup:
-    image: tecnativa/duplicity:postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}
+    image: ghcr.io/tecnativa/docker-duplicity-postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}:{{ backup_image_version }}
     hostname: backup
     {%- if macros.first_main_domain(domains_prod) %}
     domainname: {{ macros.first_main_domain(domains_prod) }}
--- a/copier.yml
+++ b/copier.yml
@@ -414,7 +414,7 @@ smtp_relay_password:
 smtp_relay_version:
   type: str
   when: *has_smtp
-  default: "8.0"
+  default: "9.0"
   help: >-
     Doodba uses docker-mailserver as the local mail relay.
 
--- a/copier.yml
+++ b/copier.yml
@@ -411,6 +411,17 @@ smtp_relay_password:
   help: >-
     What is your SMTP password?
 
+smtp_relay_version:
+  type: str
+  when: *has_smtp
+  default: "8.0"
+  help: >-
+    Doodba uses docker-mailserver as the local mail relay.
+
+    Which version of the image do you want to use? You can check which ones are
+    available in
+    https://github.com/orgs/docker-mailserver/packages/container/docker-mailserver/versions
+
 smtp_canonical_default:
   type: str
   when: *has_smtp
@@ -452,9 +463,22 @@ backup_dst:
 
     Where should the backups be stored?
 
-backup_email_from:
+backup_image_version:
   type: str
   when: &backup_present "{{ backup_dst and true }}"
+  default: "1.0"
+  help: >-
+    Doodba uses the docker-duplicity image to make the backups.
+
+    Which version of the image do you want to use? You can check which ones are
+    available in
+    https://github.com/orgs/Tecnativa/packages/container/docker-duplicity-postgres/versions
+    and
+    https://github.com/orgs/Tecnativa/packages/container/docker-duplicity-postgres-s3/versions
+
+backup_email_from:
+  type: str
+  when: *backup_present
   help: >-
     The backup container will send email reports if the SMTP relay is properly
     configured.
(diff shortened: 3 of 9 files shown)


obayit:
I strongly approve moving images to ghcr, although for other reasons. I can't access docker hub from my location :( so I have to use VPN for every new server or for updating the images:

screenshot-hub docker com-2021 02 12-16_17_41

joao-p-marques:
> I can't access docker hub from my location :(

Yes, that is one of the reasons for migrating to GHCR :smiley:

README.md.jinja
```diff
@@ -5,6 +5,12 @@
 [![pipeline status]({{ gitlab_url }}/badges/{{ odoo_version }}/pipeline.svg)]({{ gitlab_url }}/commits/{{ odoo_version }})
 [![coverage report]({{ gitlab_url }}/badges/{{ odoo_version }}/coverage.svg)]({{ gitlab_url }}/commits/{{ odoo_version }})
 {%- endif %}
+{%- if domains_prod %}
+[![Deployment data](https://img.shields.io/badge/Deployment-prod-green)](https://{{ domains_prod[0].hosts[0] }})
```
Yajo, line 9: I think you better use the same as here: https://github.com/Tecnativa/doodba-copier-template/blob/788851afcd758e837f266867eea51a8cd46f40b2/prod.yaml.jinja#L33

OTOH, it could be `http` in some circumstances: https://github.com/Tecnativa/doodba-copier-template/blob/788851afcd758e837f266867eea51a8cd46f40b2/copier.yml#L231-L233

README.md.jinja
```diff
@@ -5,6 +5,12 @@
...
+{%- endif %}
+{%- if domains_test %}
+[![Deployment data](https://img.shields.io/badge/Deployment-demo-yellow)](https://{{ domains_test[0].hosts[0] }})
```
Yajo, line 12: Same as above, but for test:

https://github.com/Tecnativa/doodba-copier-template/blob/788851afcd758e837f266867eea51a8cd46f40b2/test.yaml.jinja#L32

https://github.com/Tecnativa/doodba-copier-template/blob/788851afcd758e837f266867eea51a8cd46f40b2/copier.yml#L281-L283

common.yaml.jinja
```diff
@@ -50,7 +50,7 @@ services:
   {%- if smtp_relay_host %}
 
   smtpreal:
-    image: tvial/docker-mailserver:latest
+    image: ghcr.io/docker-mailserver/docker-mailserver:stable
```
Yajo, line 53: Add a question to the questionary, to ask for the version you want to use here.

By default, it should be `8.0`, to avoid surprises, don't you think?

You can link to all the versions in the question if you find it useful: https://github.com/orgs/docker-mailserver/packages/container/docker-mailserver/versions

common.yaml.jinja
```diff
@@ -83,7 +83,7 @@ services:
   {%- if backup_dst %}
 
   backup:
-    image: tecnativa/duplicity:postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}
+    image: ghcr.io/tecnativa/duplicity-postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}:0.1
```
Yajo, line 86:
```suggestion
    image: ghcr.io/tecnativa/docker-duplicity-postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}:0.1
```

Yajo (changes requested):

README.md.jinja
joao-p-marques: Thanks! I hadn't seen that macro, much better now!

As we can't know if it will be `https` or not, I am leaving the link as `http`, as it will be redirected to secure if active.

common.yaml.jinja
joao-p-marques: Yes, thanks! I've added a question with that: https://github.com/Tecnativa/doodba-copier-template/compare/17b0a7860e3786ac0343c036cc96e5c812144079..6a6eb0f4cae1cb07aa03f8529550955bdf1a63c0#diff-a74d725f30a732f10530aec0b66cdc1f1856272e35ede5ba4826ead3075c3d3dR414-R424

common.yaml.jinja
joao-p-marques: :man_facepalming: MB

tests/test_settings_effect.py
```diff
@@ -47,9 +47,13 @@ def test_backup_config(
...
     if "s3" in backup_dst:
-        assert prod["services"]["backup"]["image"] == "tecnativa/duplicity:postgres-s3"
+        assert prod["services"]["backup"]["image"].startswith(
+            "ghcr.io/tecnativa/docker-duplicity-postgres-s3:"
+        )
     else:
-        assert prod["services"]["backup"]["image"] == "tecnativa/duplicity:postgres"
+        assert prod["services"]["backup"]["image"].startswith(
+            "ghcr.io/tecnativa/docker-duplicity-postgres:"
+        )
```
joao-p-marques, line 56: @Yajo WDYT about leaving this test like this? It would avoid having to change the test each time we change the docker image tag, while still checking the name is correct...

Yajo: Can't you just give an answer to the copy process and then assert that the answer is correctly used here?

docs/daily-usage.md
```diff
@@ -222,8 +222,8 @@ docker-compose -f prod.yaml up -d
 #### Backups
 
 Backups are only available in the production environment. They are provided by
-[tecnativa/duplicity:postgres-s3](https://github.com/Tecnativa/docker-duplicity). The
-structure of the backed up folder:
+[tecnativa/docker-duplicity-postgres-s3](https://github.com/Tecnativa/docker-duplicity).
```
Yajo, line 225: I think you should link to the package, or rename the link label to just `docker-duplicity`.

Yajo:
Could you please add a button to install current module also? I find myself doing that all the time via CLI.

joao-p-marques:
> Could you please add a button to install current module also? I find myself doing that all the time via CLI.

Done :heavy_check_mark: That was also something that I had in mind but never got to... Thanks for the suggestion!

image

tasks_downstream.py
```diff
@@ -244,6 +244,24 @@ def write_code_workspace_file(c, cw_path=None):
                 "problemMatcher": [],
                 "options": {"statusbar": {"label": "$(play-circle) Start Odoo"}},
             },
+            {
+                "label": "Install current module",
+                "type": "process",
+                "command": "invoke",
+                "args": ["install", "--cur-file", "${file}"],
```
Yajo, line 251: How about this?

```suggestion
                "args": ["install", "--cur-file", "${file}", "restart"],
```

So, Odoo is up and running after install?

joao-p-marques:
@Yajo tests are passing now
I just pushed again to remove "wip" from the commit message, but you can review when you have a chance :wink:

copier.yml
```diff
@@ -411,6 +411,17 @@ smtp_relay_password:
   help: >-
     What is your SMTP password?
 
+smtp_relay_version:
+  type: str
+  when: *has_smtp
+  default: "8.0"
```
Yajo, line 417: Let's update the default 😃 

```suggestion
  default: "9.0"
```

Yajo (approved):

tests/conftest.py
```diff
@@ -262,3 +266,23 @@ def some_method(self, test):
             }
         )
     build_file_tree(file_tree)
+
+
+def _containers_running(exec_path):
+    with local.cwd(exec_path):
+        if len(docker_compose("ps").splitlines()) > 2:
+            _logger.error(docker_compose("ps"))
```
Yajo, line 274: I think `docker-compose ps -q` is better for this use case.

joao-p-marques: Thanks for the suggestion, change made.
I assume you refer to the first command, right? For the log, it makes sense to keep the full `docker-compose ps` output IMHO, to help identify the container that might be causing issues.

joao-p-marques:
@Yajo I also added another small improvement with [another commit](f36abee9fdee6cda4c136d23a887755d363b6ddc), and I explain the need in the commit message. Even though the current configuration doesn't cause any issues, IMHO it is an important nice-to-have.

tests/conftest.py
Yajo: Yes, sorry.

tests/conftest.py
```diff
@@ -262,3 +266,26 @@ def some_method(self, test):
             }
         )
     build_file_tree(file_tree)
+
+
+def _containers_running(exec_path):
+    with local.cwd(exec_path):
+        if len(docker_compose("ps", "-a").splitlines()) > 0:
```
Yajo, line 273: You don't care about header lines

```suggestion
        if len(docker_compose("ps", "-aq").splitlines()) > 0:
```

Yajo (changes requested):

tests/conftest.py
joao-p-marques: Yes, sorry, MB :man_facepalming: 
I meant to use `-q`, not `-a`. Although, now that I think of it, it makes sense to use `-aq` instead of only `-q`.
Should be fixed now, along with an improved error message.

Yajo (approved):

PR #229, opened 2021-03-05
Fix pylintArgs

amon-ra:
This fix pylint path in VSCode 1.48.2

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -115,7 +115,7 @@ def write_code_workspace_file(c, cw_path=None):
             "python.linting.flake8Enabled": True,
             "python.linting.ignorePatterns": [f"{str(SRC_PATH)}/odoo/**/*.py"],
             "python.linting.pylintArgs": [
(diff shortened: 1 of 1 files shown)


joao-p-marques:
I think with this we end up with a line too long for pylint, as pointed in the tests that fail. Can you fix that @amon-ra?

OTOH, can you explain a little more why is this necessary? AFAICS, the config in the workspace file is being generated correctly:
```json
"python.linting.pylintArgs": [
      "--init-hook=\"import sys;sys.path.append('/home/jota/Documents/Tecnativa/clientes/agilar-odoo/odoo/custom/src/odoo')\"",
      "--load-plugins=pylint_odoo"
    ],
```

Yajo:
Please could you explain what problem are you fixing?

amon-ra:
Sorry for this. Pylint os not working in my setup and in first test tjis fix it. But later I realize that is has another problem.
Vscode use triple slash in configuration so I think it must be changed.

Close if it have not

PR #236, opened 2021-03-30
Improve install and test tasks

joao-p-marques:
Some improvements to the `resetdb`, `install` and `test` flows:
1. Allow to prepare DB with config (Only for v11+)
1.1. In resetdb with flag
1.2. In a separate `preparedb` task
2. Improve `test` task
2.1. Test all core, extra and private modules with one command.
2.2. Test only explicit list of modules and skip some if needed (for multi-stage testing) (Only for v12+)
3. Improve `resetdb` task:
3.1 Install all core, extra and private modules present with extra flags
3.2 Install only dependencies of any list of modules passed.
4. Improve tests
4.1. Run docker tests by default
4.2. Add tests for new additions
5. Install `click-odoo` from acsone/click-odoo-contrib#93 to fix bug

- [x] Needs https://github.com/Tecnativa/doodba/pull/407

TT27635

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -528,6 +528,31 @@ def install(c, modules=None, cur_file=None, core=False, extra=False, private=Fal
         )
 
 
+def _get_module_dependencies(c, modules=None, core=False, extra=False, private=False):
+    """Returns a list of the addons' dependencies
+
+    By default, refers to the addon from directory being worked on,
+    unless other options are specified.
+    """
+    # Get list of dependencies for addon
+    cmd = "docker-compose run --rm odoo addons list --dependencies"
+    if core:
+        cmd += " --core"
+    if extra:
+        cmd += " --extra"
+    if private:
+        cmd += " --private"
+    if modules:
+        cmd += f" -w {modules}"
+    with c.cd(str(PROJECT_ROOT)):
+        dependencies = c.run(
+            cmd,
+            env=UID_ENV,
+            hide="stdout",
+        ).stdout.splitlines()[-1]
+    return dependencies
+
+
 def _test_in_debug_mode(c, odoo_command):
     with tempfile.NamedTemporaryFile(
         mode="w", suffix=".yaml"
@@ -681,14 +706,34 @@ def stop(c, purge=False):
     help={
         "dbname": "The DB that will be DESTROYED and recreated. Default: 'devel'.",
         "modules": "Comma-separated list of modules to install. Default: 'base'.",
+        "core": "Install all core addons. Default: False",
+        "extra": "Install all extra addons. Default: False",
+        "private": "Install all private addons. Default: False",
+        "dependencies": "Install only the dependencies of the specified addons."
+        "Default: False",
     },
 )
-def resetdb(c, modules="base", dbname="devel", populate=True):
+def resetdb(
+    c,
+    modules=None,
+    core=False,
+    extra=False,
+    private=False,
+    dbname="devel",
+    populate=True,
+    dependencies=False,
+):
     """Reset the specified database with the specified modules.
 
     Uses click-odoo-initdb behind the scenes, which has a caching system that
     makes DB resets quicker. See its docs for more info.
     """
+    if dependencies:
+        modules = _get_module_dependencies(c, modules, core, extra, private)
+    elif core or extra or private:
+        modules = _get_module_list(c, modules, core, extra, private)
+    else:
+        modules = modules or "base"
     with c.cd(str(PROJECT_ROOT)):
         c.run("docker-compose stop odoo", pty=True)
         _run = "docker-compose run --rm -l traefik.enable=false odoo"
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -555,10 +555,40 @@ def _test_in_debug_mode(c, odoo_command):
         time.sleep(SERVICES_WAIT_TIME)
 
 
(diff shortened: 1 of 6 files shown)


tasks_downstream.py
```diff
@@ -651,6 +651,8 @@ def resetdb(c, modules="base", dbname="devel"):
             env=UID_ENV,
             pty=True,
         )
+        if populate:
+            c.run("preparedb", env=UID_ENV, pty=True)
```
Yajo, line 655:
```suggestion
            c.run(f"{_run} preparedb", env=UID_ENV, pty=True)
```

Yajo (changes requested):
Missing tests.

tasks_downstream.py
joao-p-marques: Change done. Thanks :+1:

tests/test_tasks_downstream.py
```diff
@@ -88,12 +106,17 @@ def test_resetdb(
...
-                stdout = invoke("resetdb")
+                stdout = invoke("resetdb", "--no-populate")
             assert "Creating database cache" in stdout
             assert "from template devel" in stdout
             assert _install_status("base") == "installed"
             assert _install_status("purchase") == "uninstalled"
             assert _install_status("sale") == "uninstalled"
+            assert not _get_config_param("report.url")
+            if supported_odoo_version >= 11:
+                stdout = invoke("resetdb", "--populate")  # (default)
```
Yajo, line 117: This way you test it's properly the default behavior.

```suggestion
                stdout = invoke("resetdb")  # --populate is the default
```

joao-p-marques:
@Yajo I just made some changes and added some new functionality, also solving your comments:

1. The **test** task now includes another 3 options:
        1. Run all tests for all OCA, private or core modules present in the environment
	2. Skip some tests of that given list
	3. Limit the tests only for that specific list being passed. Previously, if we ran `invoke test -m module1` and `module2` depended on `module1`, the tests for `module2` would also run. Now, if we are above v12, we can make use of `--test-tags` to limit that.
2. `resetdb` can now install **all** private, OCA and core modules, as well as install only its **dependencies**, in a single command
3. The `preparedb` script can now be executed in a **separate task**, and there is a proper **warning** for when the user tries to execute it in images bellow v11.

tests/test_nitpicking.py
```diff
@@ -384,7 +384,7 @@ def test_pre_commit_in_subproject(
 
 
                 class ResPartner(models.Model):
-                    _name = "res.partner"
+                    _inherit = "res.partner"
```
joao-p-marques, line 387: Just a note @Yajo. I changed this because I needed to test the actual installation of the test module, and I was having some problems with the `_name` assignment.
This shouldn't introduce any other issues, right?

Yajo: Yes, no problem.

joao-p-marques:
Blocked until https://github.com/acsone/click-odoo-contrib/pull/93 is fixed

tests/conftest.py
```diff
@@ -79,9 +88,9 @@ def cloned_template(tmp_path_factory):
 
 
 @pytest.fixture()
-def docker() -> LocalCommand:
-    if os.environ.get("DOCKER_TEST") != "1":
-        pytest.skip("Missing DOCKER_TEST=1 env variable")
+def docker(request) -> LocalCommand:
+    if request.config.getoption("--skip-docker-tests"):
+        pytest.skip("Skipping docker tests")
```
Yajo, line 93: This change probably means something must change on CI, right?

Yajo (changes requested):

tests/conftest.py
joao-p-marques: Not really. By default, this won't skip tests, only if you manually add the flag, and the invoke task will only append this flag if explicitly set:
https://github.com/Tecnativa/doodba-copier-template/pull/236/commits/ce62a143eb6717d1e7db15f19756f4fc4b478379#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R70
https://github.com/Tecnativa/doodba-copier-template/blob/stable/.github/workflows/test.yml#L79

What I can do is remove the `DOCKER_TEST` variable, as it is no longer in use :+1:

Yajo: OK, then remove the variable please.

joao-p-marques: Done :+1: Thanks!

odoo/custom/dependencies/pip.txt
```diff
@@ -1,3 +1,5 @@
 git+https://github.com/OCA/openupgradelib.git@master
 unicodecsv
 unidecode
+
+git+https://github.com/Tecnativa/click-odoo-contrib.git@fix-active-modules-hashing
```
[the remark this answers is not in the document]
Yajo: This is still missing.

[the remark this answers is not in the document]
joao-p-marques: Ok, I only added it here for testing purposes, if upstream does not answer and we merge this I will move it to the correct place and reference the PR instead of the branch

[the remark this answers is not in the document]
joao-p-marques: As it already included in Doodba with https://github.com/Tecnativa/doodba/pull/414, I'll just wait for the build and then drop this commit

Yajo (approved):

PR #240, opened 2021-04-13
Don't hardcode Jedi anymore

Yajo:
In 49cf5dbb1fe52462d84524221360b89972ffa731 we added this hardcoding because Pylance just didn't work fine.

According to my recent tests, latest versions work just fine. It's actually much faster to load file symbols.

By removing this setting from the workspace, we let each user choose the preferred language server.


joao-p-marques (approved):
Ok, nice! :+1: 
Works fine too from my testing.

PR #244, opened 2021-04-16
feat: when testing, add DB_FILTER=devel

Yajo:
Multidb environments are harder to test because any http test lands in the db selector instead of the actual instance.

The solution is to add a dbfilter, but that would become problematic in devel when doing db backups or resetdb.

Thus, when `invoke test`ing, add automatically the DB_FILTER, so there's one less thing to worry about.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -678,7 +678,7 @@ def test(
     if debugpy:
         _test_in_debug_mode(c, odoo_command)
     else:
-        cmd = ["docker-compose", "run", "--rm", "odoo"]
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):

Yajo:
Oops.. those ❌ ... time to disable old versions in CI? Just like https://github.com/Tecnativa/doodba/pull/408

joao-p-marques:
Well, maybe, although our goal was that the images would remain usable, right? Perhaps we should find a way of keeping this functionality in the template, as those kinds of projects might still be needed for migration purposes...

joao-p-marques:
@Yajo just rebased from `devel`. Let's see if the tests now pass...

PR #246, opened 2021-04-26
Update from oca/maintainer-quality-tools to oca/oca-addons-repo-template

joao-p-marques:
Avoid test errors like https://github.com/Tecnativa/doodba-copier-template/runs/2408448281?check_suite_focus=true#step:13:308


joao-p-marques:
@Yajo can you review the changes to the diff files pls?

I am not sure if we should just update the diffs or fix our files accordingly instead...

Yajo (approved):

PR #247, opened 2021-04-28
Fix test task: handle test tags and modules to skip correctly

joao-p-marques:


tasks_downstream.py
```diff
@@ -651,6 +651,7 @@ def test(
         modules = cur_module
     else:
         modules = _get_module_list(c, modules, core, extra, private)
+    skip = skip or ""
```
Yajo, line 654: Just set `""` as `skip`'s default value instead of `None`

tasks_downstream.py
```diff
@@ -660,16 +661,18 @@ def test(
         raise exceptions.ParseError(
             msg="Available modes are 'init' or 'update'. See --help for details."
         )
-    odoo_command.append(modules)
     # Skip test in some modules
     modules_list = modules.split(",")
-    for m_to_skip in skip or []:
+    for m_to_skip in skip.split(","):
+        if not m_to_skip:
+            continue
```
Yajo, line 668: Passing in `--skip ,,,` should be supported? 🤔

joao-p-marques: Well, this was covering the case where `skip` is `""` (default) -> the list will be `[""]` -> the first `m_to_skip` will be `""`, which shouldn't be parsed. Is it worth preventing that specific situation and throw an error? `--skip ,,,` won't cause damage, just won't be parsed...

joao-p-marques:
@Yajo just added a new test to assert that (the skip part was not being tested)

Yajo (approved):

PR #248, opened 2021-04-28
Sync from stable

Yajo:
Just syncing from stable branch.


joao-p-marques (approved):

joao-p-marques:
Just added a commit fixing the pre-commit errors on the new issue templates. It will be added to stable later.

PR #251, opened 2021-05-05
Move db filter to docker compose + Docker-compose down instead of stop

joao-p-marques:
As mentioned in #244 (comment)
multiple DBs in an instance will cause problems in http tests
that land in the DB selector

When testing locally, the same can happend when developing controllers and APIs
When there is no session cookie, many calls fail as there is no explicit DB associated
We can limit that DB as, in devel, "devel" will be used pretty much 99% of the time
For the other cases, we can edit the DB filter manuallly

We can also remove it from the test task, as it will be included
in the docker compose file directly

---

Do a `docker-compose down` instead of a `stop`

1. When developing, this is a task generally performed when wanting to leave for another environment, not really when wanting to restart the environment relatively fast. For that, we have `restart`.

2. If regular down commands aren't executed, too many resources get accumulated (e.g. networks) so you end up having to run them manually. This would prevent that and help keep the docker environment clean in the OS.

3. There isn't a big performance impact

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -620,8 +620,6 @@ def _get_module_list(c, modules=None, core=False, extra=False, private=False):
         "cur-file": "Path to the current file."
         " Addon name will be obtained from there to run tests",
         "mode": "Mode in which tests run. Options: ['init'(default), 'update']",
-        "db_filter": "DB_FILTER regex to pass to the test container Set to ''"
-        " to disable. Default: '^devel$'",
     },
 )
 def test(
@@ -634,7 +632,6 @@ def test(
     debugpy=False,
     cur_file=None,
     mode="init",
-    db_filter="^devel$",
 ):
     """Run Odoo tests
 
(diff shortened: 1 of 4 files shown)


tasks_downstream.py
```diff
@@ -700,11 +695,9 @@ def test(
 )
 def stop(c, purge=False):
     """Stop and (optionally) purge environment."""
-    cmd = "docker-compose"
+    cmd = "docker-compose down"
```
Yajo, line 698: The resources you're cleaning up are meaningless.

Networks that are not attached consume no IP namespaces, and containers consume almost no disk space. The meaningful things are images and volumes. Besides, when pruning volumes, you may end up removing volumes that you wouldn't want to remove, because they'd be considered garbage by Docker because they'd be attached to no containers.

And stopping/starting the environment would be slower.

You'll have to do the `docker system prune` every now and then anyways due to dangling volumes and images, so what's the benefit of this? 🤔

joao-p-marques: Well, my main goal was to avoid the IP saturation caused by having so many networks...

> Networks that are not attached consume no IP namespaces

I don't know if this is true... Maybe I have something wrong with my config, but from time to time I get the error saying that docker can't find a non-overlapping IP. The networks do have containers attached, so they must be keeping some IP reference... Can you point some documentation to see if I'm doing something wrong?

That was the main benefit, and I think the slightly bigger stop/start times wouldn't be that meaningful, as oppose to the time spent each time you have to prune the networks, plus running a `docker-compose down` in each project to recreate the networks that no longer exist but that have containers associated...

Yajo (approved):

Yajo:
CI fails.

joao-p-marques:
> CI fails.

Thanks, I just need to rebase after the fix from #252

PR #252, opened 2021-05-07
Patch older builds

joao-p-marques:
`pathlib` is useful for insider and CI scripts.
As the images are frozen, we can add it here.

---

The official PG repos for Debian jessie have been archived: https://apt-archive.postgresql.org/
This adds the correct reference

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -6,6 +6,7 @@ Contains common helpers to develop using this child project.
 """
 import json
 import os
+import stat
 import tempfile
 import time
 from itertools import chain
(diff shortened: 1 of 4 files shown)


joao-p-marques:
@Yajo I moved the pip patch from https://github.com/Tecnativa/doodba-copier-template/pull/251 to here, as this are important fixes, and the others are suggestions and improvements

tasks_downstream.py
```diff
@@ -388,6 +388,10 @@ def develop(c):
     """Set up a basic development environment."""
     # Prepare environment
     Path(PROJECT_ROOT, "odoo", "auto", "addons").mkdir(parents=True, exist_ok=True)
+    if ODOO_VERSION < 11:
+        Path(PROJECT_ROOT, "odoo", "custom", "build.d", "20-update-pg-repos").chmod(
+            0o755
```
Yajo, line 393: This should move somewhere else. It doesn't make sense to keep changing permissions to this file each time we run `inv develop`.

Also, change to `0o750` if possible.

Yajo (changes requested):

tasks_downstream.py
joao-p-marques: Ok, I moved it to a copier task :+1: 

I left the mode as `0750`, as suggested. However, I see that other build files in doodba have `0755`. Could this cause problems when building in other environments?

joao-p-marques: Ok, thanks, I made the chane

PR #254, opened 2021-05-20
choices-same-type

Yajo:
This should avoid hitting https://github.com/copier-org/copier/issues/405 and will actually be an unsupported use case in Copier 6.


Yajo:
As this will mean a type change and, maybe, a migration, this is targeted for v3, so I set as draft until v2.8 is released.

joao-p-marques (approved):

Yajo:
I just tested it with https://github.com/copier-org/copier/commit/20c151e9f432987eaf75e673ac22d5a426b73774 and upgrading both in automated and interactive mode works as expected.

PR #255, opened 2021-05-21
Release v2.8.0

Yajo:


tests/test_tasks_downstream.py
```diff
@@ -226,11 +266,127 @@ def test_install_test(
                 _tests_ran(stdout, supported_odoo_version, "note")
             # Test --debugpy and wait time call with
             invoke("stop")
```
joao-p-marques, line 268: We will also need to include the changes from #254 here, this was in fact a bug

joao-p-marques (approved):

PR #259, opened 2021-06-02
fix: publisher info in comments

Yajo:
This avoids the need to remove by hand all that 🗑️.


joao-p-marques (approved):

joao-p-marques:
Pre-commit wants to reformat it..

PR #264, opened 2021-07-01
Don't include non-installable addons when retrieving list

joao-p-marques:
Give the option to include in the base function, but we won't want them by default, both when testing modules and when resetting the DB.


Yajo (approved):

Yajo:
It seems you forgot to pre-commit.

joao-p-marques:
Sorry, done now :+1:

PR #274, opened 2021-09-23
Fix Odoo list_db option

joao-p-marques:
Currently, was not doing anything


common.yaml.jinja
```diff
@@ -18,6 +18,7 @@ services:
       PGUSER: &dbuser "{{ postgres_username }}"
       DB_FILTER: "{{ odoo_dbfilter }}"
       PROXY_MODE: "{% if odoo_proxy %}true{% else %}false{% endif %}"
+      LIST_DB: "{% if odoo_listdb %}true{% else %}false{% endif %}"
```
Yajo, line 21: If you prefer it:

```suggestion
      LIST_DB: "{{ odoo_listdb | to_json }}"
```

Yajo (approved):

common.yaml.jinja
joao-p-marques: Thanks!

common.yaml.jinja
```diff
@@ -18,6 +18,7 @@ services:
       PGUSER: &dbuser "{{ postgres_username }}"
       DB_FILTER: "{{ odoo_dbfilter }}"
       PROXY_MODE: "{% if odoo_proxy %}true{% else %}false{% endif %}"
+      LIST_DB: "{{ odoo_listdb | to_json }}"
```
Yajo, line 21: Sorry, I'm already thinking on copier 6.

```suggestion
      LIST_DB: "{{ odoo_listdb | tojson }}"
```

Yajo (changes requested):

common.yaml.jinja
joao-p-marques: No problem :wink: Already changed :heavy_check_mark:

PR #277, opened 2021-09-30
fix typo in backups

ap-wtioit:


joao-p-marques (approved):

joao-p-marques:
Thanks for the correction @ap-wtioit :smiley:

Yajo (approved):

ap-wtioit:
test errors seem related to https://twitter.com/Scott_Helme/status/1443604252501417984

joao-p-marques:
Indeed... We are already looking for a fix. You can follow the process 
in <https://github.com/Tecnativa/doodba-copier-template/issues/278>

Once it is fixed, you can rebase here...

Yajo:
v10 should get out of support this week, so in case we don't get a fix in time, this could be rebased after removing that support.

joao-p-marques:
> v10 should get out of support this week, so in case we don't get a fix in time, this could be rebased after removing that support.

True :)

Anyway, once https://github.com/Tecnativa/doodba-copier-template/pull/279 is merged, you can rebase here to fix the issue to avoid the wait...

Yajo:
May you rebase please?

Issue #278, opened 2021-09-30
Error copier fails trying to build for Odoo v8: Err https://deb.nodesource.com jessie/main Sources


carlosecv:
## Describe the bug

If I try invoke img-build or invoke img-build --pull

## To Reproduce

**Affected versions**:
   Odoo v8 

run invoke img-build or invoke img-build --pull

The process stops with the present error:

Err https://deb.nodesource.com jessie/main Sources
  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
Err https://deb.nodesource.com jessie/main amd64 Packages
  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
W: Failed to fetch https://deb.nodesource.com/node_6.x/dists/jessie/main/source/Sources  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none

W: Failed to fetch https://deb.nodesource.com/node_6.x/dists/jessie/main/binary-amd64/Packages  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none

E: Some index files failed to download. They have been ignored, or old ones used instead.
Fetched 11.8 MB in 3s (3215 kB/s)
Traceback (most recent call last):
  File "/opt/odoo/common/build", line 30, in <module>
    subprocess.check_call(command)
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '/opt/odoo/custom/build.d/20-update-pg-repos' returned non-zero exit status 100
ERROR: Service 'odoo' failed to build: The command '/bin/sh -c /opt/odoo/common/build && sync' returned a non-zero code: 1

**Expected behavior** A clear and concise description of what you expected to happen.

We expect the process as no problem with certificate issue

**Additional context** Add any other context about the problem here. (e.g. OS, Docker
version, ...)

Docker version 20.10.7, build 20.10.7-0ubuntu1~20.04.1
Invoke 1.5.0
Copier _commit: v2.8.0

carlosecv:
Steps to reproduce:
copier copy gh:Tecnativa/doodba-copier-template 
answer all questions:
Odoo v8
Postgresql 9.5

[...]

PR #279, opened 2021-10-04
Fix root certificate issue for old Doodba images

joao-p-marques:
After the root LE certificate expired in October 1st, the list needs to be updated.
The alternative root certificate (ISRG Root X1) is already present as it was packaged with the old `ca-certificates`, so we just need to remove the old conflicting one to force the new one to be used.

See more at https://letsencrypt.org/certificates/

Closes https://github.com/Tecnativa/doodba-copier-template/issues/278

@Tecnativa
TT32281


Yajo (approved):
Very smart fix! 👏🏼

PR #286, opened 2021-11-04
feat: add dbfilter only to prod, defaulting to database name prefix

Yajo:
It doesn't make sense to add a DB filter to devel and test environments.

It makes sense to provide a saner default (only for production) where the filter matches any database prefixed with the name of the default database.

Benefits from https://github.com/Tecnativa/docker-duplicity/pull/75 to allow backing up only the used DBs by default.

@Tecnativa TT32631


joao-p-marques (approved):
LGTM :+1: 

If we want to disable certain DBs in devel (to avoid test errors and HTTP endpoint issues with multiple DBs) we can simply add it to the `devel` file.

PR #289, opened 2021-11-24
Allow different kinds of redirection in Traefik

joao-p-marques:
Allow to specifiy a 301 redirection instead of a 302 through a copier answer

@Tecnativa
TT31445

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -233,6 +233,7 @@ domains_prod:
         - www.shop.main.com
         - old.shop.main.com
       redirect_to: shop.main.com
+      redirect_permanent: True
 
     # VPN-only domain, with self-signed certificate
     - hosts:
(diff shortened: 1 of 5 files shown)


copier.yml
```diff
@@ -296,8 +302,12 @@ domains_test:
     - hosts: The domain names. List of strings.
     - path_prefixes: The path prefixes (if any) that will match this routing rule.
       List of strings. Default: [].
-    - redirect_to: Apply a 301 redirection to this other domain.
+    - redirect_to: Apply a redirection to this other domain. By default, it will be a
+      302 redirection, but it can be turned into a 301 redirection by enabling the
+      `redirect_301` option.
       String. Default: null.
+    - redirect_301: Return 301 instead of 302 in a Traefik redirection. Boolean.
```
Yajo, line 309: I'd say `redirect_permanent` is more understandable.

Also IMHO this should be true by default, as a proxy redirection is gonna be almost always a 301. However that could introduce some backwards incompatibility. Nothing that can't be fixed with a major release bump.

Yajo (approved):
Looks nice. Just one suggestion.

PR #291, opened 2021-12-22
fix(tasks): invoke's yaml incompatible with python 3.10

Yajo:
Invoke's vendored pyyaml is incompatible with Python 3.10, as seen in https://github.com/pyinvoke/invoke/issues/820.

As we're using it, a good workaround is to use system's un-vendored version if available.

When installing invoke with pipx, you can get the normal pyyaml with:

```sh
pipx install invoke
pipx inject invoke pyyaml
```

That + this patch fixes the problem.

Besides, I'm removing builds for py3.6, which is EOL.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -16,7 +16,11 @@ from pathlib import Path
 from shutil import which
 
 from invoke import exceptions, task
-from invoke.util import yaml
+
+try:
+    import yaml
+except ImportError:
+    from invoke.util import yaml
 
 PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.6
           - 3.7
           - 3.8
         odoo-version:


pedrobaeza:
I understand that if Python 3.6 is still used by some Odoo versions, it should be still tested, isn't it?

Yajo:
Not really, because the python versions are for DCT itself, not for the odoo that runs inside doodba. That depends on the python version specified in doodba's dockerfiles.

This should affect only the python version installed in your devs' laptops.

pedrobaeza:
OK, being Ubuntu 20.04 bundled with Python 3.8, and even Debian previous version (10) Python 3.7, is reasonable enough to remove it.

About the PR itself, not sure if there can be any compatibility in the API of yaml itself, but I can't say if such library is not very stable.

tasks_downstream.py
```diff
@@ -16,7 +16,11 @@
 from shutil import which
 
 from invoke import exceptions, task
-from invoke.util import yaml
+
+try:
+    import yaml
+except ImportError:
+    from invoke.util import yaml
```
joao-p-marques, line 23: Why not using the Python standard `yaml` library always, then? Does it break something or introduce any constraints?

joao-p-marques (commented):
LGTM overall, thanks :+1: 

Looks like CI is failing, though...

tasks_downstream.py
Yajo: Because if you install invoke with pipx, yaml won't exist inside its virtualenv unless you do the trick explained in https://github.com/Tecnativa/doodba-copier-template/pull/291#issue-1086681597. And it works for lower python versions. And in the future maybe upstream fixes https://github.com/pyinvoke/invoke/issues/820 and this doesn't become a problem when people start getting python 3.10 in their local machines.

joao-p-marques: Oh I see, I thought it came as part of the standard Python libraries. Looks good then :+1: Just missing the CI fixes.

Yajo:
May you please relaunch tests? They seem to have failed due to some timeout.

pedrobaeza:
Done.

Yajo:
So it seems ready to merge now.

joao-p-marques:
> So it seems ready to merge now.

Nice. I see you removed the 3.9 and 3.10 Python tests… I guess we should support those at some point, although it could be addressed in another PR. But what was the error you were getting?

Yajo:
Yes. DCT works on those versions, even more with this patch. But there's some problem with the CI (the failure was at install step). That goes farther than this PR scope (adding those builds was just a commodity) and can be fixed later.

joao-p-marques (approved):

joao-p-marques:
> Yes. DCT works on those versions, even more with this patch. But there's some problem with the CI (the failure was at install step). That goes farther than this PR scope (adding those builds was just a commodity) and can be fixed later.

Yes, I have also been using it with Python 3.9 and 3.10, so it works :+1: That's what I meant, let's add it in the CI later...

pedrobaeza (approved):

PR #292, opened 2021-12-23
feat: full qualified image name everywhere

Yajo:
This patch downloads the exact same images, but removes ambiguity.

When using podman as backend, you get this error with `docker-compose pull`:

    ERROR: for pgweb  failed to resolve image name: short-name resolution enforced but cannot prompt without a TTY

This is because podman supports configurable default registries to resolve ambiguous names, and when pulling an ambiguous-named image, it asks interactively how to resolve it:

```sh
podman pull sosedoff/pgweb
? Please select an image:
  ▸ registry.fedoraproject.org/sosedoff/pgweb:latest
    registry.access.redhat.com/sosedoff/pgweb:latest
    docker.io/sosedoff/pgweb:latest
    quay.io/sosedoff/pgweb:latest
```

Interactive prompts are disabled when using podman through the socket, which is the case when using docker-compose, and so it just fails.

With this patch, images are more specific, and podman works better with them out of the box.

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -46,7 +46,7 @@ services:
   {%- endif %}
 
   smtpfake:
-    image: mailhog/mailhog
+    image: docker.io/mailhog/mailhog
   {%- if smtp_relay_host %}
 
   smtpreal:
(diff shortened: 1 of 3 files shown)


joao-p-marques (approved):

joao-p-marques:
Thanks! 👍

PR #293, opened 2022-01-04
feat: experimental podman support

Yajo:
No network isolation, but the rest works OK. Basic needs:

1. Aggregate code as UID=GID=0.
2. Let subuids write to `odoo/auto`.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -24,7 +24,21 @@ except ImportError:
 
 PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
-UID_ENV = {"GID": str(os.getgid()), "UID": str(os.getuid()), "UMASK": "27"}
+UID_ENV = {
+    "GID": os.environ.get("DOODBA_GID", str(os.getgid())),
+    "UID": os.environ.get("DOODBA_UID", str(os.getuid())),
+    "DOODBA_UMASK": os.environ.get("DOODBA_UMASK", "27"),
+}
+UID_ENV.update(
+    {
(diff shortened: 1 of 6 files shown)


joao-p-marques (approved):
Looks good :+1: 

I have only tested that the plain docker usage is not affected, and so far seems ok :+1: It's nice that we can keep both :slightly_smiling_face:

tasks_downstream.py
```diff
@@ -24,7 +24,21 @@
 
 PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
-UID_ENV = {"GID": str(os.getgid()), "UID": str(os.getuid()), "UMASK": "27"}
+UID_ENV = {
+    "GID": os.environ.get("DOODBA_GID", str(os.getgid())),
```
joao-p-marques, line 28: BTW, why did you add this option? It seems ok, but I don't see you using it anywhere neither in the code nor in the documentation.

Yajo: As it was before, even if you had exported `$GID` somewhere, it wouldn't work because it would be always overriden by `os.getgid()`. Same for UID.

So I realized that it seems more safe to prefix all the env vars with `DOODBA_`, because GID, UID and UMASK are so common names that I fear they could collide somewhere else.

In docker, container UID/GID are the same, but in podman they are mapped using subuid and subgid mechanism and can easily lead to permission problems.

Yajo:
Anything here missing to merge?

joao-p-marques:
> Anything here missing to merge?

Oops. Sorry, I just didn't get around to do it yet. Let's proceed. Thanks for the PR :+1:

PR #330, opened 2022-04-07
independent wkhtmltopdf

amon-ra:
This pull try to solve Tecnativa/doodba#4039. It will load wkhtmltopdf from an external docker image in port 8080


bosd:
> I think that this should be optional. A copier question.

Indeed, That would be nice!

bosd:
Can someone re-trigger the tests?

PR #331, opened 2022-04-13
fix: remove tty from git-aggregate task

Yajo:
Without this patch, if you mispell one git remote URI, `invoke git-aggregate` will get stuck, and the only clue you'll get to know why will be a message like this buried under a lot of "success" logs:

    Username for 'https://github.com':

Now, you get a proper error because there's no TTy associated with the process, so it cannot ask for input and will produce a meaningful error.

@moduon MT-460

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -428,9 +428,8 @@ def git_aggregate(c):
     """
     with c.cd(str(PROJECT_ROOT)):
         c.run(
-            "docker-compose --file setup-devel.yaml run --rm odoo",
+            "docker-compose --file setup-devel.yaml run --rm -T odoo",
             env=UID_ENV,
-            pty=True,
         )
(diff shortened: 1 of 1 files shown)


tasks_downstream.py
```diff
@@ -428,7 +428,7 @@ def git_aggregate(c):
     """
     with c.cd(str(PROJECT_ROOT)):
         c.run(
-            "docker-compose --file setup-devel.yaml run --rm odoo",
+            "docker-compose --file setup-devel.yaml run --rm -T odoo",
             env=UID_ENV,
             pty=True,
```
Tardo, line 433: I understand that then it is not necessary to launch it with "pty" either.... no?

Yajo: Correct.

Tardo (approved):

PR #334, opened 2022-05-04
fix: update https redirection method

yajo:
Today I started noticing warnings like this in Traefik logs:

    SSLRedirect is deprecated, please use entrypoint redirection instead.

Indeed [the redirection method we were using is deprecated][1]. I changed it for [the most similar one][2].

[1]: https://doc.traefik.io/traefik/migration/v2/#headers-middleware-ssl-redirect-options
[2]: https://doc.traefik.io/traefik/middlewares/http/redirectscheme/


Tardo:
Can you check the tests?

yajo:
I think it should be OK now.

Tardo:
Thanks!

Tardo (approved):

PR #338, opened 2022-05-23
fix(podman): stop odoo before installing module

yajo:
This will make it work with podman 4 out of the box. Podman 4 still doesn't support `--link`, which is used by docker-compose when starting a parallel container. The fix is quite simple: just stop odoo before installing. The vscode task automatically calls `restart` after installation, which makes Odoo start again. Docker users will notice no difference.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -555,6 +555,7 @@ def install(
     if modules:
         cmd += f" -w {modules}"
     with c.cd(str(PROJECT_ROOT)):
+        c.run("docker-compose stop odoo")
         c.run(
             cmd,
             env=UID_ENV,


pedrobaeza:
But you may want to have Odoo running and at the same time, install a module. You'll restart later the running container. I don't like to have my other container stopped due to this reason.

yajo:
I don't usually use odoo while installing a module. Actually Odoo itself usually hangs in these situations. And even more while working with `workers=0` like we do for development. So this shouldn't harm a lot UX. Anyways, it's a technical limitation when using podman as a backend.

pedrobaeza:
No, Odoo doesn't hang on that situation, and workers=0 is independent from this. I don't see this very elegant, but it's only for development, so it's not totally critic. But isn't this happening on each possible task (update, etc)?

yajo:
Only on those that run a separate container for the task. I found the problem only in this one, I didn't really look into others deeply.

To put it simple, you can't do `docker-compose run odoo $whatever` if you did `docker-compose up -d` before. You need to stop odoo, then do the `run` and start again. All of this when podman is the backend, of course.

pedrobaeza:
That seems very restricting... OK, we can live with it for this command for now.

pedrobaeza (approved):

Tardo (approved):

PR #344, opened 2022-07-01
Imp headers project

Tardo:


tasks_downstream.py
```diff
@@ -281,6 +281,22 @@ def write_code_workspace_file(c, cw_path=None):
                     "statusbar": {"label": "$(symbol-property) Install module"}
                 },
             },
+            {
```
pedrobaeza, line 284: This is mixed, no?

Tardo: Yes, this is mixed. Sorry.

PR #355, opened 2022-10-28
[FIX] Update vscode python obsolete parameters

chienandalu:
- `python.pythonPath` is now `python.defaultInterpreterPath` (https://code.visualstudio.com/docs/python/environments#_manually-specify-an-interpreter)

cc @Tecnativa

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,15 +16,6 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
 
-  // Custom file associations
-  "files.associations": {
-    // HACK https://github.com/samuelcolvin/jinjahtml-vscode/issues/47
-    ".env.jinja": "jinja-shell"
-  },
-
-  // Local spell dictionary
-  "cSpell.words": ["odoo", "pytest", "scaffoldings", "skipif", "venv"],
-
   // General settings
   "editor.formatOnSave": true,
   "markdown.extension.toc.updateOnSave": false,
@@ -32,11 +23,9 @@
   "python.formatting.provider": "black",
   "python.linting.flake8Enabled": true,
   "python.linting.pylintEnabled": true,
-  "python.pythonPath": "${workspaceRoot}/.venv/bin/python",
-  "python.testing.nosetestsEnabled": false,
+  "python.defaultInterpreterPath": "${workspaceRoot}/.venv/bin/python",
   "python.testing.pytestArgs": ["-nauto", "tests"],
   "python.testing.pytestEnabled": true,
   "python.testing.unittestEnabled": false,
-  "restructuredtext.confPath": "",
   "search.followSymlinks": false
 }
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -138,7 +138,8 @@ def write_code_workspace_file(c, cw_path=None):
                 "--load-plugins=pylint_odoo",
             ],
             "python.linting.pylintEnabled": True,
-            "python.pythonPath": "python%s" % (2 if ODOO_VERSION < 11 else 3),
+            "python.defaultInterpreterPath": "python%s"
+            % (2 if ODOO_VERSION < 11 else 3),
             "restructuredtext.confPath": "",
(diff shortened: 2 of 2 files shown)


sergio-teruel (approved):

pedrobaeza:
I think you can't remove the file due to the test seeing the migration. You need to remove such test.

.vscode/settings.json
```diff
@@ -1,42 +0,0 @@
-{
```
yajo, line 1: This file is not meant to be removed. It's excluded from the template because it should not get copied over to projects, but it shares valid configurations for template development.

chienandalu: But isn't this file doomed to deprecation as it won't be actively mantained as nobody really uses it?

yajo: But you use it when you clone this repo...

chienandalu: Ok, I understand :sweat_smile:

pedrobaeza: But why not to rely on standard parameters for this project. Is there any customization required for it taking into account that this is only the template?

chienandalu: Fixed :)

chienandalu: > But why not to rely on standard parameters for this project. Is there any customization required for it taking into account that this is only the template?

Is just for working on the `doodba-copier-template` project itself. So it makes sense that is there.

pedrobaeza: Yes, but that's why I ask the previous questions

chienandalu: Well, they can be useful as a commonground for the doodba copier contributor, although a bunch of them are indeed outdated.

yajo (approved):

pedrobaeza:
If we let it, then we need to change https://github.com/Tecnativa/doodba-copier-template/blob/main/.vscode/settings.json#L35 as well, isn't it?

In the other hand, CI is failing, but it seems a temporary error. Let's see on the next push.

.vscode/settings.json
```diff
@@ -16,27 +16,16 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
 
-  // Custom file associations
-  "files.associations": {
-    // HACK https://github.com/samuelcolvin/jinjahtml-vscode/issues/47
-    ".env.jinja": "jinja-shell"
-  },
```
chienandalu, line 23: This issue was solved: https://github.com/samuelcolvin/jinjahtml-vscode/pull/48/files

.vscode/settings.json
```diff
@@ -16,27 +16,16 @@
...
-
-  // Local spell dictionary
-  "cSpell.words": ["odoo", "pytest", "scaffoldings", "skipif", "venv"],
```
chienandalu, line 26: This extension (streetsidesoftware.code-spell-checker) isn't in `extensions.js`

.vscode/settings.json
```diff
@@ -16,27 +16,16 @@
...
-
   // General settings
   "editor.formatOnSave": true,
   "markdown.extension.toc.updateOnSave": false,
   "python.formatting.blackArgs": [],
   "python.formatting.provider": "black",
   "python.linting.flake8Enabled": true,
   "python.linting.pylintEnabled": true,
-  "python.pythonPath": "${workspaceRoot}/.venv/bin/python",
-  "python.testing.nosetestsEnabled": false,
```
chienandalu, line 36: Not a valid setting anymore

.vscode/settings.json
```diff
@@ -16,27 +16,16 @@
...
+  "python.defaultInterpreterPath": "${workspaceRoot}/.venv/bin/python",
   "python.testing.pytestArgs": ["-nauto", "tests"],
   "python.testing.pytestEnabled": true,
   "python.testing.unittestEnabled": false,
-  "restructuredtext.confPath": "",
```
chienandalu, line 40: Not in extensions.js

chienandalu (commented):
.vscode/settings.json fixed. Some comments:

PR #359, opened 2022-11-23
support for docker compose v2

PicchiSeba:
Hi thank you for your project!

I had an issue yesterday caused by a different docker compose version that I had installed on my machine.
I Implemented a solution that takes into account the existence of both versions and picks the appropriate command.

It seems that because this is my first contribution I cannot automatically execute tests.


PicchiSeba:
Closing because it has been supported in the meantime

PicchiSeba:
Closed by mistake, sorry

tasks_downstream.py
```diff
@@ -45,6 +46,11 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
+DOCKER_COMPOSE_CMD = (
```
Tardo, line 49: I would skip this check.

Tardo (changes requested):
Please, update the readme. Thanks!

PicchiSeba:
I updated README.md with the url for docker compose (v2) installation. Do you think the version must be specified?

aleuffre:
> I updated README.md with the url for docker compose (v2) installation. Do you think the version must be specified?

I'd keep the version specified to be clear. Docker documentation always calls it Compose v2, as well.

Also, please update the "Dependencies" section of the README too.

For that bullet point, I'd write 

`- Docker with the Compose v2 plugin`

For Docker, I'd link here https://docs.docker.com/ and maybe you can point the compose link to this URL https://docs.docker.com/compose/compose-file/

PicchiSeba:
@aleuffre 
> For Docker, I'd link here https://docs.docker.com/ and maybe you can point the compose link to this URL
> https://docs.docker.com/compose/reference/

For Compose I put https://docs.docker.com/compose/ instead of the one you proposed because from there one can go directly to the download page as well the reference page you mentioned

pedrobaeza:
Please check CI.

Tardo:
Maybe it's a good time to switch to Python on Whales? https://gabrieldemarmiesse.github.io/python-on-whales/

pedrobaeza:
Do you mean in the tests? Do you know why the test fails?

PicchiSeba:
Looking at the tests I see logs like this
`/home/runner/work/doodba-copier-template/doodba-copier-template/.venv/bin/`**`docker-compose`**` run --rm -e LOG_LEVEL=WARNING -e PGDATABASE=devel odoo psql -tc 'select state from ir_module_module where name='"'"'mail'"'"''`
I guess that means we are still running tests with `docker-compose`

`from plumbum.run import docker_compose` in tests_tasks_downstream.py means that we are still using the old version.


we could write something like this instead
```py
docker_compose = local["docker compose"]
```
that way we use the old version and we can recycle the variable name

pedrobaeza:
Can you do such changes to put this green?

PicchiSeba:
```py
tests/test_tasks_downstream.py:22: in <module>
    docker_compose = local["docker compose"]
.venv/lib/python3.7/site-packages/plumbum/machines/local.py:242: in __getitem__
    return LocalCommand(self.which(cmd))
.venv/lib/python3.7/site-packages/plumbum/machines/local.py:202: in which
    raise CommandNotFound(progname, list(cls.env.path))
```
ofc it's failing. Docker Compose isn't installed in the docker image yet

PicchiSeba:
I searched the marketplace for an action with Compose V2. Could this be alright?
I chose this in particular because the author works for Docker.

.github/workflows/test.yml
```diff
@@ -56,6 +56,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install Compose
+        uses: ndeloof/install-compose-action@v0.0.1
```
Tardo, line 60: Use SHA, its more secure:

```suggestion
        uses: ndeloof/install-compose-action@c036dce35dc2a23597c03aa05f32347b2cfa0437
```

Tardo:
Yes, I meant for testing and even tasks. But better if with 'small' changes everything works correctly hehehe

Tardo (changes requested):

pedrobaeza:
There are conflicts in the branch.

PicchiSeba:
I'm sorry for the many force pushes but I'm having some trouble correctly running poetry on my machine

pedrobaeza:
It seems in the CI there are changes in the files:

https://github.com/Tecnativa/doodba-copier-template/actions/runs/4766933079/jobs/8475016255?pr=359

Can you put them correctly in your changes?

tasks_downstream.py
```diff
@@ -45,6 +45,7 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
+DOCKER_COMPOSE_CMD = "docker compose"
```
yajo, line 48: What if somebody's still using the old `docker-compose`? This would break their workflow.

Tardo: They can always set the version to one that uses docker-compose... no?

You are going to support software that is not going to be supported anymore... so there will come problems to support new versions of the docker compose manifest.

PicchiSeba:
The error seems to be in the GH Action I added to install `docker compose`.
I tried to run the exact same command on my machine and it executed flawlessly

aleuffre:
The GH action worked on most tests, just not all of them for some reason. One download returned a 403, the others executed properly

tasks_downstream.py
PicchiSeba: Originally I had implemented a way to check which version is currently installed. @Tardo suggested me to remove it completely because `docker-compose` won't be supported for much long.

yajo: I'm a bit afraid in the sense that the old `docker-compose` worked without any problems with podman.This lets me have a working machine that has no docker client or server installed.

Compose v2 is a plug-in for docker, which means you do need the docker client + plugin. Installation is more complex.

Besides, I'm currently using docker-compose v2 installed from a different package set that only provides it as a binary called `docker-compose`, avoiding the need to actually install the docker client.

Docker is banned in RHEL, Fedora and similar. Installing it there is kinda hard, while podman is available out of the box.

So, in short, if you can do a fallback, I'll be happier.

aleuffre: do you have any suggestions for how to implement such change? e.g. Environment variable, or checking if the command exists every time (like this zsh plugin https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/docker-compose/docker-compose.plugin.zsh#L2 ) or other methods?

PicchiSeba: I tried to implement a way to check which version is installed. Let's see how it works

tasks_downstream.py
```diff
@@ -45,6 +46,11 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
+DOCKER_COMPOSE_CMD = (
+    "docker compose"
+    if subprocess.getstatusoutput("docker compose")[0] == 0
+    else "docker-compose"
+)
```
yajo, line 53: You can do it simpler:

```suggestion
DOCKER_COMPOSE_CMD = shutil.which("docker-compose") or "docker compose"
```

yajo (changes requested):

tasks_downstream.py
PicchiSeba: Oh that's nice, thanks!

Quick question: shouldn't we prioritize the newer version if it's installed?

PicchiSeba: Hi @yajo , I have an issue with your suggestion.
If `shutil.which("docker-compose")` finds the path for the given command, then the old version will be used.
We could do the opposite and search if the new version is installed first, but `shutil.which` wont work, as docker compose is a subcommand for docker.

Should we still prioritize the old version?

yajo: > `shutil.which` wont work, as docker compose is a subcommand for docker

That's the reason why I suggested the old version before. You can do other more complex algorithms if you prefer, but for me this is one enough as it brings compatibility with both CLIs.

PicchiSeba:
Can we launch the pipelines again?

pedrobaeza:
It seems to fail in 3.10

Tardo:
I think that the CI is blocked due to use docker-compose (v1.29.2) and not 'docker compose' (v2.x)... any news here?

pedrobaeza:
We don't know right now how to unblock it, so if you know how, please supersede the PR and proceed.

PicchiSeba:
Superseded by #401

PR #360, opened 2022-12-02
fix(tasks): remove python hardcoded formatter

yajo:
Microsoft is moving Black support to its own extension: https://github.com/microsoft/vscode-black-formatter

There's no need to hardcode this formatter here. Besides, if you're using the Black-dedicated one, this way you get to use it also in DCT projects.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -8,6 +8,7 @@
     "firefox-devtools.vscode-firefox-debug",
     "mrorz.language-gettext",
     "ms-azuretools.vscode-docker",
+    "ms-python.black-formatter",
     "ms-python.python",
     "msjsdiag.debugger-for-chrome",
     "redhat.vscode-xml",
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -131,6 +131,7 @@ def write_code_workspace_file(c, cw_path=None):
     cw_config["settings"].update(
         {
             "python.autoComplete.extraPaths": [f"{str(SRC_PATH)}/odoo"],
+            "python.formatting.provider": "none",
             "python.linting.flake8Enabled": True,
             "python.linting.ignorePatterns": [f"{str(SRC_PATH)}/odoo/**/*.py"],
             "python.linting.pylintArgs": [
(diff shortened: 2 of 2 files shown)


pedrobaeza:
But if not indicated that way, how do you configure it automatically for the DCT projects?

yajo:
You would have to configure it for your user instead.

pedrobaeza:
Yeah, but that suboptimal, as we want a template that loads all the needed config in VSCode, so unless there's an option to put that in the project file, I don't like the change.

yajo:
Another option is to add the new black extension to recommendations and enable it here instead.

yajo:
Quoting from the extension's README:

> This extension is experimental. The plan is that it will eventually replace the black formatting functionality of [the Python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python).

So I didn't choose that option in the beginning because of this. I'm not sure everybody wants an experimental extension installed out of the box. So I decided to remove the hardcoded formatter and let the dev choose.

In light of this, what solution do you prefer?

pedrobaeza:
Yeah, but I suppose the move will be done on both sides: they remove it from Python extension and at the same time, the black extension is consolidate as the alternative not experimental, so we have 2 options:

1. Don't update till that moment happens.
2. Update already, trusting on enough stability, and we don't have to wait the exact moment.

But in any case, the move should be done IMO. Have you tested the extension to check its stability?

yajo:
I use python 3.11 so I need to use the new extension prerelease. It works fine fwiw.

pedrobaeza:
OK, then let's go with the experimental extension.

pedrobaeza (approved):

pedrobaeza:
Created pre-release for being usable from copier.

yajo:
Ok thanks!

pedrobaeza:
@yajo wasn't the pre-release version the way to not have it in copier unless forced? It seems not...

yajo:
No, you have to publish a prerelease tag according to the PEP-440 standard.

pedrobaeza:
Well, OK, thanks. Next time I will take that into account. Meanwhile, regular `copier -f update` will change this.

PR #369, opened 2023-02-15
imp: use DB_VERSION env

Tardo:
Before this pr the images always used the latest version of postgresql-client offered by default by the distro.

- Added validator in postgres_version question
- Added postgres 15
- Updated 'OLDEST_SUPPORTED_ODOO_VERSION' to 11.0
- Updated tests to use the latest postgres version
- Updated tests to use the latest copier version
- Don't deploy backup service when using psql 9.6


copier.yml
```diff
@@ -109,10 +109,6 @@ odoo_version:
   type: float
   default: 16.0
   choices:
-    - 7.0
-    - 8.0
-    - 9.0
-    - 10.0
```
yajo, line 115: I don't think this is a good solution. Even if these versions are outdated, you can always get a new customer in one of them and you'd need to scaffold it to migrate.

copier.yml
```diff
@@ -352,17 +348,19 @@ cidr_whitelist:
...
-    "10": "10"
+    "9.6 (Minimum version for Odoo >=7.0)": "9.6"
+    "10 (Minimum version for Odoo >=14.0)": "10"
     "11": "11"
-    "12": "12"
+    "12 (Minimum version for Odoo >=16.0)": "12"
     "13": "13"
-    "14": "14"
+    "14 (Maximum version for Odoo <=12.0)": "14"
+    "15": "15"
```
yajo, line 363: You can use a validator if you prefer ([new feature in Copier 6.2.0](https://github.com/copier-org/copier/blob/master/CHANGELOG.md#v620-2022-09-18)).

common.yaml.jinja
```diff
@@ -9,6 +9,7 @@ services:
     build:
       context: ./odoo
       args:
+        DB_VERSION: "{{ postgres_version or 'latest' }}"
```
yajo, line 12: To avoid failures in case this is what makes you remove support for older versions, you can always add this argument conditionally:

```suggestion
        {%- if odoo_version >= 11 %}
        DB_VERSION: "{{ postgres_version or 'latest' }}"
        {%- endif %}
```

yajo (changes requested):

copier.yml
Tardo: I don't know what I'm doing wrong, but I can't get the "validator" to work... :/

copier.yml
```diff
@@ -352,17 +352,24 @@ cidr_whitelist:
...
+    "11": null
+    "12": null
+    "13": null
+    "14": null
+    "15": null
+  validator: >-
+    {% if (odoo_version >= 16.0 and postgres_version < 12) or (odoo_version >= 14.0 and
+    postgres_version < 10) %} The postgres version is too low for the selected Odoo
+    version. {% elif odoo_version <= 12.0 and postgres_version > 14 %} The postgres
+    version is too high for the selected Odoo version. {% endif %}
```
yajo, line 372: This seems like won't work. All values are `null`, but they should be strings, just like before. Also, since they are strings, you need to cast them when doing the version comparison.

pyproject.toml
```diff
@@ -6,9 +6,9 @@ name = "doodba-copier-template"
 version = "0.1.0"
 
 [tool.poetry.dependencies]
-copier = "^5.1.0"
+copier = "^6.2.0"
```
yajo, line 9:
```suggestion
copier = ">=7.0.0"
```

copier.yml
```diff
@@ -10,7 +10,7 @@ _envops:
   variable_start_string: "{{"
 
 # Other Copier configurations
-_min_copier_version: "4.1.0"
+_min_copier_version: "6.2.0"
```
yajo, line 13: Now that you're updating, do it more 😋 

```suggestion
_min_copier_version: "7.0.0"
```

copier.yml
Tardo: Umm I still continue without get this to work... :(

Tardo:
Ok, I have some "problems" to get this PR ready:
- A new release must be made in docker-duplicity, currently the tests are using the master/edge version.
- I can't get the copier validator to work. I have tried the simplest way: "{% if True %}Error{% endif %}" or directly "Error" with no results.

** See the first comment for more information about the changes made.

yajo:
> * I can't get the copier validator to work

Maybe because validators only work interactively? Validators on CLI/API calls are implemented in https://github.com/copier-org/copier/pull/962, which is not yet released.

pedrobaeza:
I have just made a release of docker-duplicity

copier.yml
```diff
@@ -352,9 +352,10 @@ cidr_whitelist:
     ⚠️ It must be a list. And this is only supported if you deploy with Traefik 2+.
 
 postgres_version:
-  default: "14"
+  default: "15"
   help: >-
-    Which PostgreSQL version do you want to deploy?
+    Which PostgreSQL version do you want to deploy? (Recommended: for new instances
+    always use the lastest version. Version 9.6 has no backup support.)
```
pedrobaeza, line 358:
```suggestion
    Which PostgreSQL version do you want to deploy? (Recommended: for new instances,
    always use the latest version. Version 9.6 has no backup support.)
```

tests/test_migrations.py
```diff
@@ -76,13 +82,14 @@ def test_transtion_to_copier(
...
             assert dep_file.read_text().endswith("\n# a comment")
+        # FIXME: Maybe can't delete old files due they don't have a trace?
         # Check migrations ran fine
-        assert not (tmp_path / ".travis.yml").exists()
-        assert not (tmp_path / ".vscode" / "doodba").exists()
-        assert not (tmp_path / ".vscode" / "doodbasetup.py").exists()
-        assert not (
-            tmp_path / "odoo" / "custom" / "src" / "private" / ".empty"
-        ).exists()
+        # assert not (tmp_path / ".travis.yml").exists()
```
pedrobaeza, line 87: Why is this commented?

Tardo: I don't know the real reason for these tests but they fail with the new versions of copier. I have left a "FIXME" comment.

PR #373, opened 2023-03-08
feat: add module readme generator

yajo:
This allows doodba private modules to benefit from having a generated readme.

The template is pretty similar to the one in https://github.com/OCA/maintainer-tools, but it removes all OCA-specific stuff, making it more suitable for doodba.

@moduon MT-956 https://github.com/OCA/maintainer-tools/pull/557


yajo:
Not in this case, as this template is almost completely different to OCA's. However there were some tests failing that should be fixed now.

yajo:
The CI failure is not related to this PR

pedrobaeza:
Not sure what can happen. Can you give me any idea?

yajo:
I opened https://github.com/Tecnativa/doodba-copier-template/issues/374 to keep both issues focused 😊

yajo:
I see that #375 went ✅  so maybe it was just a temporary failure. I rebased; let's see...

yajo:
Coool, it's green now!

PR #376, opened 2023-03-15
fix(podman): allow downloading correct backups

yajo:
`domainname` is not supported when the docker backend is podman. It is required to compute appropriately the prefixes in s3 flavored images.

With this patch, we make the container compatible with both docker and podman.

@moduon MT-2568

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -84,10 +84,7 @@ services:
 
   backup:
     image: ghcr.io/tecnativa/docker-duplicity-postgres{% if backup_dst.startswith(('boto3+s3://', 's3://', 's3+http://')) %}-s3{% endif %}:{{ backup_image_version }}
-    hostname: backup
-    {%- if macros.first_main_domain(domains_prod) %}
-    domainname: {{ macros.first_main_domain(domains_prod) }}
-    {%- endif %}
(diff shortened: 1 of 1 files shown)


Tardo (approved):

pedrobaeza:
I see domainname in other places like https://github.com/Tecnativa/doodba-copier-template/pull/376/files#diff-787ea6fbd147b5cef0f0d03fd414753668838aae3648588a3ef9d22006e50ce3R54

Is this retrocompatible with running instances? I mean: doing `up -d`, the container is correctly replaced and no side effects?

yajo:
> I see domainname in other places like https://github.com/Tecnativa/doodba-copier-template/pull/376/files#diff-787ea6fbd147b5cef0f0d03fd414753668838aae3648588a3ef9d22006e50ce3R54. Didn't they need to be replaced as well?

It could be done, although it isn't a big problem unless you run production on podman.

> Is this retrocompatible with running instances? I mean: doing `up -d`, the container is correctly replaced and no side effects?

Yes, it's the same, but expressed in different ways.

pedrobaeza:
> It could be done, although it isn't a big problem unless you run production on podman.

Isn't this what you want to do, and thus, it's better to make it in one shot?

yajo:
I only use podman for dev ATM, but there you have the change anyways.

PR #378, opened 2023-04-12
imp: use docker compose v2

Tardo:
From the end of June 2023 Compose V1 won’t be supported anymore and will be removed from all Docker Desktop versions.

docker-compose = v1
docker compose = v2


Tardo:
Done in https://github.com/Tecnativa/doodba-copier-template/pull/359

PR #379, opened 2023-04-19
fix: safer merge conflict commit protection

yajo:
This is the configuration recommended upstream: https://copier.readthedocs.io/en/stable/updating/#preventing-commit-of-merge-conflicts

It is more important now that Copier supports inline conflict markers on updates.

@moduon MT-2626

### DIFF
--- a/.module-readme.rst.j2
+++ b/.module-readme.rst.j2
@@ -48,7 +48,8 @@
 {{- fragment('HISTORY', 'Changelog') -}}
 
 Credits
-=======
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Is the CI error something transient?

Putting this option makes it compatible with older versions of copier?

yajo:
Hah! There was a tricky edge case. Solved now.

yajo:
Added a workaround for https://github.com/pre-commit/pre-commit-hooks/issues/985.

pedrobaeza (approved):

PR #380, opened 2023-04-20
fix: don't recommend sass-indented extension

yajo:
Odoo uses scss. This extension only adds support for sass. Thus, it's useless to recommend it.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -12,7 +12,6 @@
     "ms-python.python",
     "msjsdiag.debugger-for-chrome",
     "redhat.vscode-xml",
-    "samuelcolvin.jinjahtml",
-    "syler.sass-indented"
+    "samuelcolvin.jinjahtml"
   ]
 }


pedrobaeza:
Sass was sorted and used in some Odoo versions, so I think the recommendation should remain for that Odoo versions.

chienandalu:
I'd say we can move on without it :sweat_smile: . For older versions (was it v11 the latest one?) you can still install it yourself. Recommended extensions should stick to the state of the art IMO

pedrobaeza:
Are the extensions activated by workspace? If so, I think we should still keep it by version, being this a template that can change according selecting answers.

yajo:
These are just recommendations. It doesn't mean you'll always install all, or that they'll be enabled. That's up to the dev. Thus, I agree with @chienandalu.

pedrobaeza:
OK, let's remove it, but the CI is red. for P3.9, v15

pedrobaeza:
Relaunched to see if now all goes well and it was a transient error.

PR #390, opened 2023-06-23
imp: support traefik 3.x

Tardo:
Follow: https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3/

- Updated poetry dependencies
- Updated pre-commit dependencies
- Use copier 8.0.0 = updated tests

** Now all (non-required) 'str' questions has a default value ("")... due to new copier don't support NoneType.
** New copier question "traefik_version"
** Now rendered compose manifests only include traefik labels for the selected version
** Fixed the latest version of postgresql that can be used in Odoo 11.0 and 12.0


.pre-commit-config.yaml.jinja
```diff
@@ -104,40 +103,45 @@ repos:
...
         name: flake8 only __init__.py
         args: ["--extend-ignore=F401"] # ignore unused imports in __init__.py
         files: /__init__\.py$
-        additional_dependencies: ["flake8-bugbear==20.1.4", "importlib-metadata<5.0.0"]
+        additional_dependencies: ["flake8-bugbear==23.3.12", "importlib-metadata==4.2.0"]
+      - id: flake8
+        name: flake8 only __manifest__.py
+        args: ["--extend-ignore=B018"] # ignore useless expressions in __manifest__.py
+        files: /__manifest__\.py$
+        additional_dependencies: ["flake8-bugbear==23.3.12", "importlib-metadata==4.2.0"]
```
aleuffre, line 121: Instead of having the double hook, what do you think about having a single hook and adding at the end of `.flake8` 

```
per-file-ignores=
    __init__.py:F401
    __manifest__.py:B018
```
?

PR #391, opened 2023-06-29
Update README for Copier v8 compatibility

HekkiMelody:
Simple documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,16 @@ pipx ensurepath
 Once you installed everything, you can now use Copier to copy this template:
 
 ```bash
-copier copy gh:Tecnativa/doodba-copier-template ~/path/to/your/subproject
+copier copy --UNSAFE gh:Tecnativa/doodba-copier-template ~/path/to/your/subproject
 ```
 
 Copier will ask you a lot of questions. Answer them to properly generate the template.
+
+The `--UNSAFE` flag is necessary since
+[Copier v8](https://copier.readthedocs.io/en/stable/configuring/#unsafe) to use features
+such as Jinja extensions, Migrations and Tasks.
+
+Make sure you trust the code you're executing!
 
 Notes:
 
@@ -83,13 +89,13 @@ If you always used Copier with this project, getting last updates with Copier is
 
 ```bash
 cd ~/path/to/your/downstream/scaffolding
-copier update
+copier update --UNSAFE
 ```
 
 Copier will ask you all questions again, but default values will be those you answered
 last time. Just hit <kbd>Enter</kbd> to accept those defaults, or change them if
-needed... or you can use `copier --force update` instead to avoid answering again all
-things.
+needed... or you can use `copier update --force --UNSAFE` instead to avoid answering
+everything again.
 
 Basically, read Copier docs and `copier --help-all` to know how to use it.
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't better to adapt templates to a "safe" syntax?

aleuffre:
> Isn't better to adapt templates to a "safe" syntax?

Probably. To be honest, I didn't look into it and wouldn't know how, and since the project heavily relies on all three of "Jinja extensions, Migrations and Tasks", I assumed it would require a heavy heavy refactor at the very least.

pedrobaeza:
@yajo as you know both pieces. Are there too many changes to do?

README.md
```diff
@@ -83,13 +89,13 @@ If you always used Copier with this project, getting last updates with Copier is
...
 cd ~/path/to/your/downstream/scaffolding
-copier update
+copier update --UNSAFE
 ```
 
 Copier will ask you all questions again, but default values will be those you answered
 last time. Just hit <kbd>Enter</kbd> to accept those defaults, or change them if
-needed... or you can use `copier --force update` instead to avoid answering again all
-things.
+needed... or you can use `copier --force update --UNSAFE` instead to avoid answering
```
yajo, line 97: In copier 8+, all flags are *after* the subcommand. This has changed for ergonomy, and because there's a new subcommand that required a different set of flags. 

Thus, this has changed too:

```suggestion
needed... or you can use `copier update --UNSAFE --force` instead to avoid answering
```

README.md
```diff
@@ -64,11 +64,17 @@ pipx ensurepath
...
 
 ```bash
-copier copy gh:Tecnativa/doodba-copier-template ~/path/to/your/subproject
+copier copy --UNSAFE gh:Tecnativa/doodba-copier-template ~/path/to/your/subproject
 ```
 
 Copier will ask you a lot of questions. Answer them to properly generate the template.
 
+The `--UNSAFE` flag is necessary since
+[Copier v8](https://copier.readthedocs.io/en/latest/configuring/#unsafe) to use features
```
yajo, line 73: I think most people will use stable copier releases:

```suggestion
[Copier v8](https://copier.readthedocs.io/en/stable/configuring/#unsafe) to use features
```

yajo (commented):
You can bump this too if you want, to avoid a warning. But that doesn't affect the rest of the template: https://github.com/Tecnativa/doodba-copier-template/blob/9e48836230b19afc25103776f86077b3c5037b40/copier.yml#L13

Unsafe features here are tasks and migrations. But migrations only block updates that actually use them, so the surface of unsafety is lower. You can remove all tasks and then you won't require `--UNSAFE` in most cases: https://github.com/Tecnativa/doodba-copier-template/blob/9e48836230b19afc25103776f86077b3c5037b40/copier.yml#L43C1-L45

However that can impact the UX of the template, so it's up to you to decide what you prefer.

FWIW in the next release there'll be a `--trust` flag that is equivalent to `--UNSAFE`. Some people will prefer it because it's less scary. But that's not released yet.

aleuffre:
Should I continue simply updating the documentation, or make the necessary changes so that it doesn't use "unsafe" features anymore?

Tardo:
I updated here the tests/copier template to use copier 8.0.0 here: https://github.com/Tecnativa/doodba-copier-template/pull/390

aleuffre:
> I updated here the tests/copier template to use copier 8.0.0: #390

Nice work! That matches the documentation update I was suggesting.

PR #401, opened 2023-08-23
Support for docker compose v2

Tardo:
Supersede #359

1. Tests uses "python on whales" (only docker compose >= v2): https://gabrieldemarmiesse.github.io/python-on-whales/
2. task_downstream.py can use docker-compose or 'docker compose'.
3. Python 3.7 was dropped from github tests and added python 3.11
4. Due to dependencies needs python >= 3.8.1
5. Updated pre-commit packages and dependencies (template project & rendered project)
6. Copier >= 8.0.0 doesn't support NoneType in strings questions... so now all not required str questions have an empty string as default value (The str to None issue causes that when launching 'copier update' on projects with "None" responses, recent versions of copier throw an exception. ~The only solution I have seen is to manually edit the response file to have an empty string instead of None)~.
7. Due to the changes made here 'https://github.com/Tecnativa/doodba-copier-template/commit/3224dc34c4a675bcf10a57c6c9d7b8e9d1093ce7#diff-0f76a5aa80b9c2ce9f078384ac3e4bccae3defa883c7183632932e624765de51', I understand that the regex that excludes files has started to work. This means that if there is only one test module, once ignored it will not be able to leave this state. To solve this, a second test module is generated (installable=False).
8. ~The latest version of pylint-odoo is not used because it no longer supports many of the configured checks.~
9. Removed unsupported migration tests
10. Updated "faq" doc
11. Updated tests mqt diff samples and removed unused
12. Updated hostnames to ensure be always unique


Tardo:
ping @yajo @pedrobaeza @PicchiSeba 

Look at the first message, especially to see what you think of points 6 and 7

Tardo:
I think that i can do somehting with copier "migration" tasks to ensure that the answers has valid values. Also, now that I understand this process I can get the 'test_migrations.py' tests ready again.

About point 7. What happens is that if you only have one module and in the manifest you change it to "installable=false" the pre-commit task 'oca-update-pre-commit-excluded-addons' puts it in the ignore list in pre-commit. So, if you put that module back to "installable=True", since it is ignored... it will not run the 'oca-update-pre-commit-excluded-addons' task again and it will not be removed from the ignore list.
The moment you modify something outside the module, the 'oca-update-pre-commit-excluded-addons' task can again detect that it is as "installable=True" and remove it from the list.

pedrobaeza:
OK about point 6, although I think that it should be handled at copier level to accept NoneType / empty string for backwards compatibility and don't require migration scripts.

About point 7, we do that on new Odoo version with custom modules, and when we switch them back to installable=True, they are removed from the list with pre-commit, so I don't think there's any problem there. Maybe you are not doing `pre-commit run -a`, and thus, they are not being considered.

yajo:
> * Copier >= 8.0.0 doesn't support NoneType in strings questions... so now all not required str questions have an empty string as default value (The str to None issue causes that when launching 'copier update' on projects with "None" responses, recent versions of copier throw an exception. The only solution I have seen is to manually edit the response file to have an empty string instead of None).

Correct. You can also [recopy](https://copier.readthedocs.io/en/stable/reference/cli/#copier.cli.CopierRecopySubApp) the template.


> * Due to the changes made here '[3224dc3#diff-0f76a5aa80b9c2ce9f078384ac3e4bccae3defa883c7183632932e624765de51](https://github.com/Tecnativa/doodba-copier-template/commit/3224dc34c4a675bcf10a57c6c9d7b8e9d1093ce7#diff-0f76a5aa80b9c2ce9f078384ac3e4bccae3defa883c7183632932e624765de51)', I understand that the regex that excludes files has started to work. This means that if there is only one test module, once ignored it will not be able to leave this state. To solve this, a second test module is generated (installable=False).

Correct.

> * The latest version of pylint-odoo is not used because it no longer supports many of the configured checks.

OK. You could also update the configurations. Take a look at https://github.com/OCA/oca-addons-repo-template/pull/204 for inspiration.

Tardo:
@pedrobaeza @yajo The problems it is giving is because of the diffs with the oca-addons-repo-template repo. I don't really know the reason for testing this, I guess to detect changes. If you see the changes correct I will update the diffs.
From what I see in 'oca-addons-repo-template' the 'pylintrc*' files use deprecated pylint-odoo directives for the version used in Odoo 16.0.
The changes I have applied here differ quite a bit from those applied in the repo.

yajo:
> I don't really know the reason for testing this, I guess to detect changes

Yes, that's it basically.

Back then when I designed that test, it was to assume that we have different configs, but that's the "supported diff".

When upstream evolves, we should update and, if it makes sense, change that diff.

.pre-commit-config.yaml.jinja
```diff
@@ -44,7 +57,7 @@ repos:
           - --branch={{ odoo_version }}
           - --template-filename=.module-readme.rst.j2
   - repo: https://github.com/myint/autoflake
-    rev: v1.4
+    rev: v2.2.0
```
pedrobaeza, line 60: Should this (and the rest of libraries) be unconditionally changed for all Odoo versions?

.pre-commit-config.yaml.jinja
```diff
@@ -104,40 +115,45 @@ repos:
...
         args:
-          - --valid_odoo_versions={{ "%.1f"|format(odoo_version) }}
           - --rcfile=.pylintrc-mandatory
+          - odoo/auto/addons
+        verbose: true
         additional_dependencies:
-          - isort==4.3.21
-          - pylint-odoo==5.0.5
+          - isort=={{ proj_rev.isort }}
+          - pylint-odoo=={{ proj_rev.pylint_odoo }}
```
aleuffre, line 154: what do you think about following OCA and using the pylint_odoo hook directly?

I've had trouble in the past with trying to update this hook and ending up with incompatible versions of pylint and pylint_odoo. Would also simplify the file; the isort dependency is not really necessary since we're running the isort hook separately too.

I understand if it's out of scope though.

.pre-commit-config.yaml.jinja
aleuffre: I've personally been updating the hook versions in this file even on projects still using odoo 12, and have not come across any particular issues when doing so. I vote in favour.

.pylintrc-mandatory.jinja
```diff
@@ -3,19 +3,41 @@ load-plugins=pylint_odoo
...
 readme_template_url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
 manifest_required_keys=license
 manifest_deprecated_keys=active
 valid_odoo_versions={{ odoo_version }}
+{%- else %}
+readme-template-url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
+manifest-required-keys=license
+manifest-deprecated-keys=active
+valid-odoo-versions={{ odoo_version }}
+{%- endif %}
```
aleuffre, line 16: I don't think this depends on the odoo_version. This really depends on the pylint_odoo version, since at some point they changed the keys from having underscores (old version) to having dashes (new version) to better follow conventions

aleuffre:
haven't had time to look at all the changes, but just wanted to say, thank you so much for your work :pray:

.pre-commit-config.yaml.jinja
pedrobaeza: OK then

.pylintrc-mandatory.jinja
Tardo: The problem is that the pylint-odoo version is not known at this point. But you are right, it would be clearer if it was compared with the pylint-odoo version.
In the end, the pylint-odoo version used depends on the Odoo version. So, it can be used if you know which version you are using.

.pre-commit-config.yaml.jinja
Tardo: done!

Tardo:
Ok, all green.

tasks_downstream.py
```diff
@@ -1,11 +1,12 @@
 """Doodba child project tasks.
 
-This file is to be executed with https://www.pyinvoke.org/ in Python 3.6+.
+This file is to be executed with https://www.pyinvoke.org/ in Python 3.8.1+.
```
pedrobaeza, line 3: Is a requirement P3.8? There's still Doodba versions with lower Python versions?

pedrobaeza (commented):
Thanks for the huge effort. Some comments inline.

tasks_downstream.py
yajo: One thing ar dev tools (such as invoke), and other thing is Odoo (which runs inside the container, with a specified python version). Py3.6 is EOL since long ago.

pedrobaeza: OK, got it.

common.yaml.jinja
```diff
@@ -54,7 +54,7 @@ services:
 
   smtpreal:
     image: ghcr.io/docker-mailserver/docker-mailserver:{{ smtp_relay_version }}
-    domainname: "{{ smtp_canonical_default }}"
+    hostname: "smtp.{{ smtp_canonical_default }}"
```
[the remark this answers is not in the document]
Tardo: From what I gather from the documentation, hostname must be used: https://docker-mailserver.github.io/docker-mailserver/latest/examples/tutorials/basic-installation/

But yes, it is not necessary to hardcode the service. It should not really affect nothing since the mail service is only used for outgoing mail... and odoo uses the network alias created in 'prod.yaml'.

I have also updated the rest of the hostnames. No one should be affected by these changes.

.pre-commit-config.yaml.jinja
```diff
@@ -1,6 +1,16 @@
 {# We cannot use the same pre-commit config as MQT's because paths don't match
    and it's designed only for Odoo 13.0+, so we design our own version, which
    should be as close as possible to MQT's -#}
+
+{#- This namespace holds repo versions #}
+{%- set proj_rev = namespace() %}
+
+{%- if odoo_version < 16 %}
+  {%- set proj_rev.pylint_odoo = "v8.0.0" %}
```
bosd, line 9: Why is this pinned to version 8.0.0?
Scrolling trough changes notes of pylint_odoo I did'nt see any breaking changes.

I'm succesfully using version 8.0.20 for odoo V14

Tardo: The available checks was changed... if you run pylint-odoo in verbose mode you will see warnings messages.

pedrobaeza: Let's keep it for now in that version. You can make a PR updating to that version and changing the checks names.

pedrobaeza (approved):
Let's go with this for grenifying the CI and having the new docker compose specification. Thanks Alex for the work.

pedrobaeza:
@Yajo @Tardo I'm applying this new template over a v16 project and I'm finding a big diff in the pylint checks:

```diff
-enable=anomalous-backslash-in-string,
-    api-one-deprecated,
-    api-one-multi-together,
-    assignment-from-none,
-    attribute-deprecated,
-    class-camelcase,
-    dangerous-default-value,
-    dangerous-view-replace-wo-priority,
-    duplicate-id-csv,
-    duplicate-key,
-    duplicate-xml-fields,
-    duplicate-xml-record-id,
-    eval-referenced,
-    eval-used,
-    incoherent-interpreter-exec-perm,
+# messages that do not cause the lint step to fail
+enable=attribute-deprecated,
+    consider-merging-classes-inherited,
+    invalid-commit,
     license-allowed,
     manifest-author-string,
     manifest-deprecated-key,
@@ -45,41 +34,39 @@ enable=anomalous-backslash-in-string,
     method-inverse,
     method-required-super,
     method-search,
-    missing-import-error,
-    missing-manifest-dependency,
-    openerp-exception-warning,
-    pointless-statement,
-    pointless-string-statement,
+    missing-readme,
+    odoo-addons-relative-import,
     print-used,
-    redundant-keyword-arg,
-    redundant-modulename-xml,
-    reimported,
-    relative-import,
-    return-in-init,
-    rst-syntax-error,
     sql-injection,
-    too-few-format-args,
     translation-field,
     translation-required,
-    unreachable,
     use-vim-comment,
-    wrong-tabs-instead-of-spaces,
-    xml-syntax-error,
-    # messages that do not cause the lint step to fail
-    consider-merging-classes-inherited,
-    create-user-wo-reset-password,
-    dangerous-filter-wo-user,
-    deprecated-module,
-    file-not-used,
-    invalid-commit,
-    missing-newline-extrafiles,
-    missing-readme,
-    no-utf8-coding-comment,
-    odoo-addons-relative-import,
-    old-api7-method-defined,
-    redefined-builtin,
-    too-complex,
-    unnecessary-utf8-coding-comment
+    attribute-string-redundant,
+    bad-builtin-groupby,
+    context-overridden,
+    development-status-allowed,
+    except-pass,
+    external-request-timeout,
+    manifest-data-duplicated,
+    manifest-maintainers-list,
+    missing-return,
+    no-raise-unlink,
+    no-wizard-in-models,
+    no-write-in-compute,
+    odoo-exception-warning,
+    renamed-field-parameter,
+    resource-not-exist,
+    test-folder-imported,
+    translation-contains-variable,
+    translation-format-interpolation,
+    translation-format-truncated,
+    translation-fstring-interpolation,
+    translation-not-lazy,
+    translation-positional-used,
+    translation-too-few-args,
+    translation-too-many-args,
+    translation-unsupported-format,
+    website-manifest-key-not-valid-uri
```

Do you know why?

Tardo:
@pedrobaeza this is due a change in the available checks. I don't understand how such a big change is treated as a minor change... but pylint-odoo throws warnings if you keep using the usual checks :/

Readme v8.0.20 (Odoo >= 16.0): https://github.com/OCA/pylint-odoo/tree/v8.0.20
Readme v8.0.0 (Odoo <= 15.0): https://github.com/OCA/pylint-odoo/tree/v8.0.0

Tardo:
@pedrobaeza and the version of pylint-odoo used for Odoo >16.0 comes from here: https://github.com/OCA/oca-addons-repo-template/blob/41a1360e877407db3800ee66afc5bbf192f23c89/src/.pre-commit-config.yaml.jinja#L52

pedrobaeza:
I think the key is that they are now checked through https://github.com/OCA/odoo-pre-commit-hooks, but its inclusion in the pre-commit config file is missing, unlike in OCA template:

https://github.com/OCA/oca-addons-repo-template/blob/41a1360e877407db3800ee66afc5bbf192f23c89/src/.pre-commit-config.yaml.jinja#L116C34-L116C55

Can you please include such hook?

PR #413, opened 2023-10-31
[FIX] The SMTP domain should prefix `smtp`

pedrobaeza:
If not, mails are not delivered.

@Tecnativa

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -51,7 +51,7 @@ services:
 
   smtpreal:
     image: ghcr.io/docker-mailserver/docker-mailserver:{{ smtp_relay_version }}
-    hostname: "{{ smtp_canonical_default }}"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Well, I'm not experiencing the same, so better to let it as it was, unless you have any side-effect that let you to change this.

Tardo:
No problem. The only thing is the issue of having the 'smtp' part hardcoded... but I don't see a major problem either.

Maybe are matching 'smtp_canonical_default' with the hostname of the odoo service?

pedrobaeza:
Yes, I think it has to be with the docker internal naming scheme, not being able to route from  `odoo` container to `smtp` container. As I'm not sure how to fix it, let's go to the hardcoded thing.

Tardo:
Ok, my apologies for my mistake

pedrobaeza:
No worries. It has been detected soon enough.

PR #433, opened 2023-11-28
Upgrade112023

syci:


pedrobaeza:
@syci why not proposing things upstream instead of forking? Anyway, you can't simply overlap our credits in the README, as you are hijacking our work.

pedrobaeza:
cc @omar7r please don't hijack our credits in your fork.

PR #425, opened 2023-11-11
[ADD] Add option for Odoo 17.0

celm1990:
I'm not sure if it's a good idea to use Copier 9, but I have problems with the old version. @yajo, can you please review?

TODO: adapt github actions to V17 and PY 3.10

Is need execute` copier copy` with option `--trust ` similar to https://github.com/OCA/oca-addons-repo-template/pull/202

PR #434, opened 2023-12-05
[IMP] Add .ruff_cache to .gitignore

PabloEForgeFlow:
https://github.com/Tecnativa/doodba-copier-template/pull/425 replaced most linters with `ruff` based on https://github.com/OCA/oca-addons-repo-template/commit/49b350fd795e039963e153943a8d0635be6f0fee, but didn't ignore `/.ruff_cache`.


pedrobaeza:
Shouldn't be such cache on the personal profile instead of each repo?

PabloEForgeFlow:
I think so as well, but for some reason the [default path](https://docs.astral.sh/ruff/settings/#cache-dir) is the root of the current project.

pedrobaeza:
I think we should set `RUFF_CACHE_DIR` in the pre-commit environment somehow. How is this done in OCA?

PabloEForgeFlow:
We could do that by adding `cache-dir = "~/.cache/ruff"` to `.ruff.toml`. I simply decided to follow what they did as per https://github.com/OCA/oca-addons-repo-template/commit/49b350fd795e039963e153943a8d0635be6f0fee.

pedrobaeza (approved):

PR #438, opened 2023-12-19
[IMP] Set default dockermailserver to v13.0

josep-tecnativa:
CC @Tecnativa TT46621


pedrobaeza (approved):
Tested on real environment and working

PR #440, opened 2023-12-20
fix(ruff): less pedantic max-line-length config

yajo:
Until template v5.x, we just used Black, which didn't cry about line
length. After an upgrade to the latest template, there are a lot of E501
errors that are not really helpful. The autoformatter will fix most
cases, but sometimes having a longer line is more readable.

Here, I'm increasing the limit to 120 characters. If it's still too low,
we can just remove E501 from the list of errors.

@moduon MT-4544 @emiliopascual

### DIFF
--- a/.ruff.toml.jinja
+++ b/.ruff.toml.jinja
@@ -1,4 +1,10 @@
-target-version = "py310"
+target-version = {% if odoo_version < 14 -%}
+"py37"
+{%- elif odoo_version < 16 %}
+"py38"
+{%- else %}
+"py310"
+{%- endif %}
 fix = true
 cache-dir = "~/.cache/ruff"
 
--- a/.ruff.toml.jinja
+++ b/.ruff.toml.jinja
@@ -12,6 +12,9 @@ extend-select = [
 ]
 exclude = ["setup/*"]
 
+[lint.pycodestyle]
+max-line-length = 120
+
 [format]
 exclude = ["setup/*"]
 


yajo:
Docs about this change: https://docs.astral.sh/ruff/settings/#pycodestyle-max-line-length

EmilioPascual (approved):

pedrobaeza:
I prefer not to do that. The limit should be the same 88 cols.

yajo:
Until ruff was introduced, we were using black. Quoting from [their docs](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length):

> If you’re paid by the line of code you write, you can pass --line-length with a lower number. Black will try to respect that. However, sometimes it won’t be able to without breaking other rules. In those rare cases, auto-formatted code will exceed your allotted limit.

Thus we just accepted that case where longer lines can exist here and there and all is fine.

Why this change of mind now?

yajo:
No, we didn't do that here.

pedrobaeza:
That's an anomaly from OCA, so I prefer to stick to it (in fact, I thought it was activated, so that wasn't done on purpose), and as you have said, most of the times, black does its job, and you won't find too many errors due to this new linter. Other new linters may be more annoying than this one and you have needed to adapt them (like removing string from tree), so let's keep this good practice on length.

yajo:
Ok, nevermind, we'll configure this on our own repos.

PR #441, opened 2023-12-21
fix(ruff): consider odoo version to establish python target version

yajo:
@moduon MT-4544

These versions match the ones in the dockerfiles of doodba. Well, some of them are older than 3.7, but ruff doesn't support that, so by putting py37 on those, we get closer to something useful.

### DIFF
--- a/.ruff.toml.jinja
+++ b/.ruff.toml.jinja
@@ -1,4 +1,4 @@
-target-version = "py310"
+target-version = {% if odoo_version < 14 %}"py37"{% elif odoo_version < 16 %}"py38"{% else %}"py310"{% endif %}
 fix = true
 cache-dir = "~/.cache/ruff"
 


.ruff.toml.jinja
```diff
@@ -1,4 +1,10 @@
-target-version = "py310"
+target-version = {% if odoo_version < 14 -%}
```
pedrobaeza, line 1: Shouldn't all this be in the same line?

yajo: I just thought it'd be more readable like this.

I added `-` to the jinja markers, to make sure it strips whitespace.

pedrobaeza: It still seems incorrect.

yajo: I changed it to one line.

pedrobaeza (approved):

PR #455, opened 2024-03-22
[ADD] Add option to use traefik 3. 

josep-tecnativa:
With traefik 3 selected, domains rule has to be modified, in order to not use syntax like this:
`
PathPrefix(`/url1/`, `/url2/`)`, and use instead `PathPrefix(`/url1/`) || PathPrefix(`/url2/`)` because firts syntax like, is not supported by traefik 3

CC @Tecnativa TT48394


pedrobaeza:
@Tardo if you are going to be at OCA days, both can join efforts in this.

pedrobaeza:
I suppose this is already done, isn't it?

PR #457, opened 2024-03-26
[UPD] poetry + docker compose + use string-templates

Tardo:
Try to fix tests...

PR #458, opened 2024-03-26
fix(prettier): rollback to working version

yajo:
As you can see reported in https://github.com/prettier/prettier/issues/15696, prettier 3.x won't work with plugins under the context of pre-commit. In practice, this means for us that XML won't get auto-formatted.

Pinning latest working combination.

@moduon MT-5455


pedrobaeza:
#457 is maybe the solution to the broken tests?

yajo:
Let's see. I rebased.

PR #464, opened 2024-04-10
 [FIX] Correct always using docker compose, even if it not compatible.

josep-tecnativa:
In this change https://github.com/Tecnativa/doodba-copier-template/commit/e2bc946fa20edc34e9007c40c3d41e0491710fca is always used docker compose. In docker versions 24.0.6 or lower, this is incompatible, and needs to use docker-compose

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -8,6 +8,7 @@ import json
 import os
 import shutil
 import stat
+import subprocess
 import tempfile
 import time
 from datetime import datetime
@@ -46,8 +47,17 @@ ODOO_VERSION = float(
         "build"
     ]["args"]["ODOO_VERSION"]
 )
-DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" or shutil.which(
-    "docker-compose"
+# Depending on the user's docker version either version of docker compose could not
+# be available. We default to v2 and fallback to v1.
+
+docker_compose_v2 = (
+    subprocess.run([shutil.which("docker"), "compose"], capture_output=True).returncode
+    == 0
+)
+DOCKER_COMPOSE_CMD = (
+    f"{shutil.which('docker')} compose"
+    if docker_compose_v2
+    else shutil.which("docker-compose")
 )
 
 _logger = getLogger(__name__)


tasks_downstream.py
```diff
@@ -46,9 +46,14 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
-DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" or shutil.which(
-    "docker-compose"
+
+docker_compose_result = subprocess.run(
+    [which("docker"), "compose"], capture_output=True, text=True
```
chienandalu, line 51: This parameter isn't needed really
```suggestion
    [which("docker"), "compose"], capture_output=True
```

tasks_downstream.py
```diff
@@ -46,9 +46,14 @@
...
 )
```
chienandalu, line 52:
```suggestion
docker_compose_v2 = subprocess.run(
    [which("docker"), "compose"], capture_output=True
).returncode == 0
```

tasks_downstream.py
```diff
@@ -46,9 +46,14 @@
...
+if docker_compose_result.returncode == 0:
+    DOCKER_COMPOSE_CMD = "docker compose"
+else:
+    DOCKER_COMPOSE_CMD = "docker-compose"
```
chienandalu, line 56: You need the whole path to run the commands
```suggestion
DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" if docker_compose_v2 0 else  shutil.which(
    "docker-compose"
```

chienandalu (changes requested):

tasks_downstream.py
```diff
@@ -46,8 +47,15 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
-DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" or shutil.which(
-    "docker-compose"
+
+docker_compose_v2 = (
+    subprocess.run([which("docker"), "compose"], capture_output=True).returncode == 0
```
chienandalu, line 52: Don't take my word for granted :sweat_smile: 
```suggestion
    subprocess.run([shutil.which("docker"), "compose"], capture_output=True).returncode == 0
```

tasks_downstream.py
```diff
@@ -46,8 +47,15 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
-DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" or shutil.which(
-    "docker-compose"
+
```
chienandalu, line 50:
```suggestion
# Document the code here :smile: 
```

tasks_downstream.py
```diff
@@ -46,8 +47,15 @@
...
+docker_compose_v2 = (
+    subprocess.run([which("docker"), "compose"], capture_output=True).returncode == 0
+)
+
```
chienandalu, line 54:
```suggestion
```

chienandalu (changes requested):

tasks_downstream.py
```diff
@@ -46,8 +47,17 @@
         "build"
     ]["args"]["ODOO_VERSION"]
 )
-DOCKER_COMPOSE_CMD = f"{shutil.which('docker')} compose" or shutil.which(
-    "docker-compose"
+# The following lines of code determine the docker dompose command to be used.
+# The logic behind this is that if it's a version compatible with cocker compose,
+# it will use this; otherwise, it will fallback to using docker-compose instead.
```
chienandalu, line 52: Be more concise :) Also explain why
```suggestion
# Depending on the user's docker version either version of docker compose could not be available. We default to v2 and fallback to v1.
```

chienandalu (changes requested):
The commit message isn't very helpful. What about:

[FIX] docker compose command fallback

chienandalu (approved):
:+1: Although now tests are red :thinking:

josep-tecnativa:
Seems like the odd CI error is due to the line being too long, but no error is triggered in the local `pre-commit run -a`.

PR #471, opened 2024-05-18
[IMP] New parameter 'tags' to enable filtering tests according to tags.

celm1990:
When tags are specified, the `module_list` is ignored because Odoo executes tests with an OR condition, meaning all tests will be executed. See examples below.

Execute only test with tags `MyTag`
`invoke test -m MyModule --tags=MyTag`

Execute only a function called test_my_custom(Note use . to referral to function) 
`invoke test -m MyModule --tags=.test_my_custom`

More examples
https://www.odoo.com/documentation/17.0/es/developer/reference/backend/testing.html#examples

TODO:

- [ ] add test

@pedrobaeza could you take a look, what do you think?

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -748,6 +748,7 @@ def test(
     cur_file=None,
     mode="init",
     db_filter="^devel$",
+    tags=None,
 ):
     """Run Odoo tests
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@josep-tecnativa can you check why the red CI?

PR #475, opened 2024-06-04
[IMP] doodba-copier-template: domains_prod and domains_test

norlinhenrik:
Answering the copier questions, I was misleaded by the `domain_prod` and `domain_test` default value `{}` when a custom value should be a list.

I suggest default value `null`. When the user deletes the default value, a simple placeholder will show up.

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -214,8 +214,12 @@ gitlab_url:
 domains_prod:
   type: yaml
   multiline: true
-  default: {}
+  default: null
   placeholder: |
+    - hosts:
+        - www.main.com
+        - shop.main.com
+  examples: |
     # Main domains, with TLS and cert resolved with "letsencrypt" resolver
     - hosts:
         - www.main.com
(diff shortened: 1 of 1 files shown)


copier.yml
```diff
@@ -214,8 +214,12 @@ gitlab_url:
 domains_prod:
   type: yaml
   multiline: true
-  default: {}
+  default: null
   placeholder: |
+    - hosts:
```
pedrobaeza, line 219: I don't see the need of this, as there's already the same 3 lines below. Maybe we should add `cert_resolver: letsencypt` by default. cc @josep-tecnativa

norlinhenrik: I have removed duplicate lines.
The placeholder has the comment in the end, to show first the syntax of the value.

norlinhenrik:
The `inverseproxy_shared` network was not created out of the box. So I've updated the documetation to say that the traefik yaml should be inside an `inverseproxy` directory.

@pedrobaeza @josep-tecnativa Is there anything you would like me to change in this PR?

Tardo:
This was a change I made due to problems with "None" values. But if with "null" everything works correctly, please change the rest of the defaults to null... it would be the closest step to how it was before my changes: https://github.com/Tecnativa/doodba-copier-template/blame/9e48836230b19afc25103776f86077b3c5037b40/copier.yml

Docs: https://github.com/Tecnativa/doodba-copier-template/blob/main/docs/faq.md#when-upgrading-from-an-old-template-copier-fails-with-invalid-answer-none-how-to-update

** And yes, I made a mistake in considering those questions as dictionaries and not as a list of dictionaries.

PR #476, opened 2024-06-13
[IMP] Traeffik 3 compatibility

josep-tecnativa:
New Feature Added: Doodba Copier Template Now Compatible with Traefik 3

With these changes, the Doodba Copier Template is now compatible with Traefik 3. When selecting Traefik as the proxy, you can choose which version to use. By default, version 2 is selected as it is the most stable.

Additionally, this PR includes the Traefik version in the test matrix and excludes tests with insecure connections for Traefik 3 due to incompatibilities.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,10 @@ jobs:
             odoo-version: 17.0
           - python-version: 3.11
             odoo-version: 17.0
+        traefik_version:
+          - 1
+          - 2
+          - 3
 
     steps:
       # Shared steps
@@ -92,7 +96,9 @@ jobs:
       - run: poetry run invoke test
         env:
           SELECTED_ODOO_VERSIONS: ${{ matrix.odoo-version }}
+          TRAEFIK_VERSION: ${{ matrix.traefik_version }}
       # Concurrent tests (isolated)
       - run: poetry run invoke test --sequential
         env:
           SELECTED_ODOO_VERSIONS: ${{ matrix.odoo-version }}
+          TRAEFIK_VERSION: ${{ matrix.traefik_version }}
--- a/test.yaml.jinja
+++ b/test.yaml.jinja
@@ -1,6 +1,7 @@
 {%- import "_macros.jinja" as macros -%}
 {%- import "_traefik1_labels.yml.jinja" as traefik1_labels -%}
 {%- import "_traefik2_labels.yml.jinja" as traefik2_labels -%}
+{%- import "_traefik3_labels.yml.jinja" as traefik3_labels -%}
 {%- set _key = traefik2_labels.key(project_name, odoo_version, "test") -%}
 version: "2.4"
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,6 @@ from typing import Dict, Union
 
 import pytest
 import yaml
-from packaging import version
 from plumbum import ProcessExecutionError, local
 from plumbum.cmd import git, invoke
 from python_on_whales import DockerClient
@@ -24,6 +23,8 @@ with open("copier.yml") as copier_fd:
 # Different tests test different Odoo versions
 OLDEST_SUPPORTED_ODOO_VERSION = 11.0
 ALL_ODOO_VERSIONS = tuple(COPIER_SETTINGS["odoo_version"]["choices"])
+TRAEFIK_VERSION = os.getenv("TRAEFIK_VERSION", "3")
+
 SUPPORTED_ODOO_VERSIONS = tuple(
     v for v in ALL_ODOO_VERSIONS if v >= OLDEST_SUPPORTED_ODOO_VERSION
 )
@@ -67,9 +68,6 @@ DBVER_PER_ODOO = {
         "latest": LATEST_PSQL_VER,
     },
 }
-
-# Traefik versions matrix
-ALL_TRAEFIK_VERSIONS = ("latest", "1.7")
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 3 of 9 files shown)


tests/conftest.py
```diff
@@ -24,6 +23,9 @@
 # Different tests test different Odoo versions
 OLDEST_SUPPORTED_ODOO_VERSION = 11.0
 ALL_ODOO_VERSIONS = tuple(COPIER_SETTINGS["odoo_version"]["choices"])
+# ALL_TRAEFIK_VERSIONS = tuple(COPIER_SETTINGS["traefik_version"]["choices"])
```
pedrobaeza, line 26: Remove commented code

tests/test_routing.py
```diff
@@ -66,6 +67,8 @@ def test_multiple_domains(
     if supported_odoo_version < 16:
         data["postgres_version"] = 13
     dc = DockerClient(compose_files=[f"{environment}.yaml"])
+    # Guarda la configuración en una variable como una cadena JSON
```
pedrobaeza, line 70: In English, please, but is it still relevant?

tests/test_routing.py
```diff
@@ -66,6 +67,8 @@ def test_multiple_domains(
...
+    # docker_compose_config_json = json.dumps(docker_compose_config, indent=2)
```
pedrobaeza, line 71: Remove commented code

_traefik3_2_labels.yml.jinja
pedrobaeza, line 1: Why this file name? Shouldn't it be `_traefik_3_labels_2.yml.jinja`? And why not everything together?

josep-tecnativa: No, the name cannot be _traefik_3_labels_2.yml because the different templates later ignored like this:
https://github.com/Tecnativa/doodba-copier-template/blob/main/copier.yml#L8
And if I use that name, they wouldn't match the Regex. Anyway, I have changed the name  to _traefik3_paths_labels.yml.jinja since this part of the template is responsible for completing the rules with Path and PathPrefix. And it is done with two templates because that is how it works, for example, with Traefik 2, although in that case, DCT performs the first part with the Traefik 1 template since it is compatible, and the second with the Traefik 2 template. I have followed this logic, although with the changes in Traefik 3, for the first part, I also had to create a specific template for Traefik 3.

.github/workflows/test.yml
```diff
@@ -49,6 +49,10 @@ jobs:
             odoo-version: 17.0
           - python-version: 3.11
             odoo-version: 17.0
+        traefik_version:
```
pedrobaeza, line 52: When updating an old template, which value will take by default?

pleirb: I have tried both cases, using an old template and creating a new one. In both cases there do not seem to be any big changes regarding the Trafic labels defined for the Odoo service in prod.yaml. In fact it doesn't let me choose the Traefik version

imagen

josep-tecnativa: Hi @pleirb , to update a template using the changes from this PR, use the command `copier update --vcs-ref fix-ci-error --trust`. Regarding what you mentioned, @pedrobaeza , by default in an old template Traefik 2 will be selected. In fact, as it has been done until now, there is no real difference between selecting Traefik 1 or Traefik 2. This is how it has worked so far.

pedrobaeza (approved):

.github/workflows/test.yml
pleirb: > Hola .pleirb , para actualizar una plantilla usando los cambios de esta PR, utilice el comando `copier update --vcs-ref fix-ci-error --trust`. Con respecto a lo que ha mencionado, se @pedrobaezaseleccionará por defecto en una plantilla antigua Traefik 2. De hecho, como se ha hecho hasta ahora, no hay diferencia real entre la selección de Traefik 1 o Traefik 2. Así ha funcionado hasta ahora.

I took the following steps to work with this PR

```
git clone https://github.com/Tecnativa/doodba-copier-template.git
cd doodba-copier-template
git fetch origin pull/476/head:pr-476
git checkout pr-476
cd ..
copier copy ./doodba-copier-template ./project-16-prod --trust
```
I answered all the questions as I normally do and then applied the command you suggested.
```
cd project-16-prod
copier update --vcs-ref fix-ci-error --trust
```
But I am getting the following error:
`Destination repository is dirty; cannot continue. Please commit or stash your local changes and retry.`
I have tried with a new project and with one already in use. In both cases I get this error. What will I be doing wrong?

pleirb:
Be careful, the version that was published for Traefik 3 does not support exposed databases or subdomains.

imagen

I hope the tests carried out here are helpful: [Solution](https://github.com/Tecnativa/doodba-copier-template/discussions/478#discussioncomment-9930620)

celm1990:
Hi, I tested Traefik 3, but the [paths_without_crawlers](https://github.com/Tecnativa/doodba-copier-template/blob/4c33dd2f09dab4426dc605536848d0430705a857/copier.yml#L337) feature is not working. If you navigate to `/website/info`, it still shows the information. Is this change intentional?

josep-tecnativa:
> Hi, I tested Traefik 3, but the [paths_without_crawlers](https://github.com/Tecnativa/doodba-copier-template/blob/4c33dd2f09dab4426dc605536848d0430705a857/copier.yml#L337) feature is not working. If you navigate to `/website/info`, it still shows the information. Is this change intentional?

It was a mistake. Fixed here https://github.com/Tecnativa/doodba-copier-template/pull/483

test.yaml.jinja
```diff
@@ -1,6 +1,7 @@
 {%- import "_macros.jinja" as macros -%}
 {%- import "_traefik1_labels.yml.jinja" as traefik1_labels -%}
 {%- import "_traefik2_labels.yml.jinja" as traefik2_labels -%}
+{%- import "_traefik3_labels.yml.jinja" as traefik3_labels -%}
```
celm1990, line 4: @josep-tecnativa  could you please confirm if, for test domains, labels are generated correctly? I see this import, but it is not used.

pedrobaeza: He's on holidays till the second week of September.

PR #479, opened 2024-07-03
[ADD] Feature to expose database

josep-tecnativa:
Since Traefik 3 is avaliable on Doodba Copier Template, we can configure with our template, a postgres exposed database. This is only working on Traefik 3.

- [x] Correct when serveral domains, HostSNI duplicated map key traefik label.

### DIFF
--- a/_traefik3_paths_labels.yml.jinja
+++ b/_traefik3_paths_labels.yml.jinja
@@ -42,6 +42,16 @@
   {%- endif %}
 {%- endmacro %}
 
+{%- macro domains_rule_sni(domain_group) -%}
+  {%- for host in domain_group.hosts -%}
+    HostSNI(`{{ host }}`)
+    {%- if not loop.last -%}
+      ||
+    {%- endif -%}
+  {%- endfor -%}
+{%- endmacro %}
+
+
 {%- macro key(project_name, odoo_version, suffix) %}
   {{- '%s-%.1f-%s'|format(project_name, odoo_version, suffix)|replace('.', '-') }}
 {%- endmacro %}
@@ -210,7 +220,10 @@
 
 {%- macro database(domain_groups_list, cidr_whitelist, key, port, project_name) %}
       {#- Service #}
+      traefik.tcp.routers.{{ key }}-database.entrypoints: postgres-entrypoint
       traefik.tcp.services.{{ key }}-database.loadbalancer.server.port: 5432
+      traefik.tcp.routers.{{ key }}-database.tls: "true"
+      traefik.tcp.routers.{{ key }}-database.tls.certResolver: letsencrypt
 
       {%- if cidr_whitelist %}
       {#- Declare whitelist middleware #}
@@ -220,23 +233,12 @@
         {%- endfor %}
       {%- endif %}
 
-      {%- call(domain_group) macros.domains_loop_grouped(domain_groups_list) %}
+      {#- Apply rule to the first element in domain_groups_list #}
+      {%- set first_domain_group = domain_groups_list[0] %}
+      traefik.tcp.routers.{{ key }}-database.rule: {{ domains_rule_sni(first_domain_group) }}
       {#- Remember basic middlewares for this domain group #}
       {%- set _ns = namespace(basic_middlewares=[]) -%}
       {%- if cidr_whitelist %}
         {%- set _ns.basic_middlewares = _ns.basic_middlewares + ["whitelist"] %}
       {%- endif %}
-
-      {#- database router #}
-      {{-
-        router_tcp(
-          domain_group=domain_group,
-          key=key,
-          suffix="database",
-          service="database",
-          middlewares=_ns.basic_middlewares,
-          port=port,
-        )
-      }}
-      {%- endcall %}
 {%- endmacro %}
--- a/prod.yaml.jinja
+++ b/prod.yaml.jinja
@@ -94,7 +94,7 @@ services:
     labels:
       traefik.enable: "true"
       traefik.docker.network: "inverseproxy_shared"
-      {{- traefik2_labels.database(
+      {{- traefik3_labels_2.database(
         domains_prod,
         postgres_cidr_whitelist,
         _key,
--- a/prod.yaml.jinja
+++ b/prod.yaml.jinja
@@ -88,6 +88,7 @@ services:
       - .docker/db-creation.env
     restart: unless-stopped
     {%- if postgres_exposed %}
+    {%- if traefik_version == 3 %}
     networks:
       default:
       inverseproxy_shared:
@@ -101,6 +102,10 @@ services:
         postgres_exposed_port,
         project_name,
       ) }}
+    {%- else %}
+    ports:
+      - "{{ postgres_exposed_port }}:5432"
+    {%- endif %}
     {%- endif %}
   {%- endif %}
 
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Can you rebase?

prod.yaml.jinja
```diff
@@ -94,7 +94,7 @@ services:
     labels:
       traefik.enable: "true"
       traefik.docker.network: "inverseproxy_shared"
-      {{- traefik2_labels.database(
+      {{- traefik3_paths_labels.database(
```
celm1990, line 97:
```suggestion
      {{- traefik3_labels_2.database(
```
I get this error because [import](https://github.com/Tecnativa/doodba-copier-template/blob/e653c6e0de09f00368dc85040df92d2852134982/prod.yaml.jinja#L7) is imported as alias `traefik3_labels_2`
image

image

celm1990 (changes requested):

prod.yaml.jinja
josep-tecnativa: Thank you @celm1990

celm1990 (approved):
LGTM, thanks a lot for improving this feature

josep-tecnativa:
ping @pedrobaeza

pedrobaeza (commented):
As this doesn't work in Traefik 2-, you can put for them the port direct mapping instead of limiting the question to Traefik 3

prod.yaml.jinja
```diff
@@ -102,6 +103,19 @@ services:
...
+   db:
+    extends:
+      file: common.yaml
+      service: db
+    env_file:
+      - .docker/db-creation.env
+    restart: unless-stopped
+    {%- if postgres_exposed %}
+    ports:
+      - "{{ exposed_port }}:5432"
```
pedrobaeza, line 116: Why repeating some part of the structure? And this should be only if port exposed, not on every case.

pedrobaeza:
It seems OK in code review, but you should try to generate a scaffolding on each type to be sure, or add them to tests.

yajo (changes requested):
Sorry, I don't mean to sound rude, but I have the feeling all this is wrong... 😅 

- TCP routers don't have a concept of "path".
- Are you sure TLS termination _can_ be done in Traefik for a TCP router?
- Are you seriously exposing the database _always, by default and with no opt-out option_?

josep-tecnativa:
> Sorry, I don't mean to sound rude, but I have the feeling all this is wrong... 😅
> 
>     * TCP routers don't have a concept of "path".
> 
>     * Are you sure TLS termination _can_ be done in Traefik for a TCP router?
> 
>     * Are you seriously exposing the database _always, by default and with no opt-out option_?

First of all, thanks for pointing out that TCP routers don't have the concept of a path. I have already corrected that.

Secondly, I would say that Traefik 3 can handle TLS termination for TCP routers. This functionality is documented in the official Traefik documentation [here](https://doc.traefik.io/traefik/routing/routers/#hostsni-and-hostsniregexp). But please correct me if I'm wrong, although I believe it is correct.

Lastly, no, by default I do not expose the database. It is only exposed if the option to expose it is chosen, as you can see here:
https://github.com/Tecnativa/doodba-copier-template/blob/081f73791dd284f0687be63205cd1dceb80e18f5/prod.yaml.jinja#L90

_traefik3_paths_labels.yml.jinja
```diff
@@ -210,7 +220,11 @@
 
 {%- macro database(domain_groups_list, cidr_whitelist, key, port, project_name) %}
       {#- Service #}
+      traefik.tcp.routers.{{ key }}-database.entrypoints: postgres-entrypoint
       traefik.tcp.services.{{ key }}-database.loadbalancer.server.port: 5432
+      traefik.{{ key }}-database.port: 5432
```
yajo, line 225: This one looks like a traefik v1 label

yajo (commented):
Thanks for attending the previous comment! Yes, it seems I was wrong in those last 2 points.

I see this also would need some changes on the traefik deployment, so you probably need to update the docs, as now there must exist a `postgres-entrypoint`.

Thank you

josep-tecnativa:
This is ready to merge @pedrobaeza .

pedrobaeza:
@yajo is it OK for you?

_traefik3_paths_labels.yml.jinja
josep-tecnativa: Done. Thank you for the review.

prod.yaml.jinja
```diff
@@ -88,19 +88,24 @@ services:
       - .docker/db-creation.env
     restart: unless-stopped
     {%- if postgres_exposed %}
+    {%- if traefik_version == 3 %}
```
yajo, line 91: AFAICS, this should work with traefik 2 too: https://doc.traefik.io/traefik/v2.11/routing/routers/#rule_1

```suggestion
    {%- if traefik_version > 1 %}
```

yajo (changes requested):

prod.yaml.jinja
josep-tecnativa: The problem with this is that Traefik 2 cannot differentiate between multiple databases by domain. You would need to create an entrypoint and open a port for each database within a server. I have tested this quite a few times.

So, we could say that it does not work in the most optimal way. That’s why I think it should currently be configured only for Traefik 3, and later we could add support for Traefik 2 and specify its limitations.

I’m currently working on the test.

yajo: Whoa, that's weird because it claims to support also the HostSNI router. But well, my experience is based on reading the docs; if yours is based on experimentation, then I'll trust you! Thanks for sharing.

prod.yaml.jinja
```diff
@@ -88,19 +88,24 @@ services:
...
+      {{- traefik3_labels_2.database(
         domains_prod,
         postgres_cidr_whitelist,
         _key,
         postgres_exposed_port,
         project_name,
       ) }}
+    {%- else %}
+    ports:
+      - "{{ postgres_exposed_port }}:5432"
```
yajo, line 107: `postgres_exposed_port` should be a conditional question, only asked if using traefik 3 or newer.

yajo (changes requested):

prod.yaml.jinja
josep-tecnativa: I added this because with Traefik 1 and 2, direct access to the database can be achieved by exposing the port and making some additional configuration in PostgreSQL. Although it's not too difficult to set the ports manually, I believe this should be set up to generate automatically if needed.

yajo: Yes, I just mean that Copier shouldn't ask what port you want to expose if you're running on traefik 3 and thus you're not gonna expose it.

josep-tecnativa: You're right, I didn't understand you correctly. Thanks!

josep-tecnativa:
@ap-wtioit Do you have any idea why I can't connect to the database in the new test?

https://github.com/Tecnativa/doodba-copier-template/actions/runs/11326688958/job/31496138294#step:16:279

josep-tecnativa:
Since Docker does not allow connections through ports on the same host starting from version 25, it is currently not possible to create a test. The functionality of this PR has been tested by both @celm1990  and me. Therefore, I believe it is ready to be merged (once CI finishes), and if in the future I find a way to create the test, I will implement it.

ping @pedrobaeza

pedrobaeza (approved):

PR #482, opened 2024-07-09
[FIX] Traefik 2 multiple main hosts

josep-tecnativa:
CC @Tecnativa TT49975


josep-tecnativa:
In this PR, the Doodba Copier Template has been adapted to support multiple main domains in Traefik 2. This change has been made because previously, the syntax of Traefik 1 was used.

josep-tecnativa:
Could you review @pedrobaeza ?

pedrobaeza (approved):
Trusting on the real tests done.

PR #483, opened 2024-07-23
[FIX] Without forbidden crawlers

josep-tecnativa:
Removed the conditional that prevented the Traefik 3 feature without forbidden crawlers, which was added by mistake.


josep-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):

PR #485, opened 2024-07-25
[FIX] Delete obsolete labels

josep-tecnativa:
CC @Tecnativa TT50321


josep-tecnativa:
@yajo please, check this out

_traefik2_hosts_labels.yml.jinja
```diff
@@ -51,20 +51,6 @@
...
-        router(
-          prefix="forbiddenCrawlers",
-          index0=domain_group.loop.index0,
-          rule=domains_rule(domain_group.hosts, paths_without_crawlers),
-          entrypoints=domain_group.entrypoints,
-        )
-      }}
-      {%- endif %}
-
       {#- Normal routers #}
```
yajo, line 54: These routers are wrong too.

yajo (changes requested):
Since you're dropping support for traefik v1, you should remove that option from the questionnaire.

As a side note, you might consider adding [CTT](https://github.com/KyleKing/copier-template-tester) to your pipeline, so you can have a snapshots-based rendering of the project that helps you spot these failures early. I do it [here](https://gitlab.com/moduon/odoo-module-template) and it helps a lot.

pedrobaeza:
We mustn't remove Traefik 1 option. Just don't populate its labels if Traefik 2 is selected. And isn't this file exclusive for Traefik 2?

PR #490, opened 2024-09-25
[FIX] Pre-commit problems for odoo versions with python < 3.7

josep-tecnativa:
CC  @Tecnativa TT50985
As ruff is not compatible with Python versions below 3.7, the pre-commit is causing problems in Odoo versions below 13.0.

- [x] Test pre-commit on odoo < 13.0 scaffolding
- [x] Test pre-commit on odoo >= 13.0 scaffolding to make sure that it  keeps using ruff instead of flake8

### DIFF
--- a/.flake8.jinja
+++ b/.flake8.jinja
@@ -1,0 +1,12 @@
+[flake8]
+max-line-length = 88
+max-complexity = 16
+# B = bugbear
+# B9 = bugbear opinionated (incl line length)
+select = C,E,F,W,B,B9
+# E203: whitespace before ':' (black behaviour)
+# E501: flake8 line length (covered by bugbear B950)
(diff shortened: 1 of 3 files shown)


josep-tecnativa:
This is ready to merge @pedrobaeza

pedrobaeza (approved):

.flake8.jinja
celm1990, line 1: Can it be rendered only when necessary?.

The file name can be(the same for _isort.cfg_)

` {% if odoo_version <= 12 %}.flake8{% endif %}.jinja`

See https://copier.readthedocs.io/en/stable/configuring/#conditional-files-and-directories for more information

josep-tecnativa: Thank you!

PR #491, opened 2024-09-26
[IMP] Render flake and isort only if necessary

josep-tecnativa:
As mentioned here https://github.com/Tecnativa/doodba-copier-template/pull/490#discussion_r1776036501


josep-tecnativa:
ping @pedrobaeza

PR #496, opened 2024-10-18
[ADD] Older versions compatibility.

josep-tecnativa:
based on https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja#L96


.pre-commit-config.yaml.jinja
```diff
@@ -69,47 +104,49 @@ repos:
       - id: ruff-format
   {% else -%}
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 20.8b1
```
pedrobaeza, line 107: Why changing this and the others in the file? Isn't better to create a new file for those needing these versions?

josep-tecnativa: I have changed those versions, adapting them to ones that are compatible with Odoo versions lower than 13.0.

Regarding your suggestion to use different files, this is not possible because the project already has a pre-commit-config.yaml to manage the hooks for the doodba-copier-template. If we have a single pre-commit-config.yaml.jinja file, when the scaffolding is generated, it overwrites the pre-commit-config.yaml. However, if I create files like {% if odoo_version >= 13 %}pre-commit-config.yaml{% endif %}.jinja, as you mentioned, we end up with the pre-commit-config.yaml from the doodba-copier-template, which is not what we want.

PR #500, opened 2024-10-25
[DCK] Add all domains to external database connection

josep-tecnativa:
Updated the template to include all domains in the external database access rule, allowing users to access the database from any configured domain.


josep-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):

PR #501, opened 2024-10-25
Update faq: update --force is deprecated

bosd:
as of https://github.com/copier-org/copier/commit/6dde037423da28940abeed081d978af621c4ac38 the `--force` flag is removed. 

It results in: Error: Unknown switch --force

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -408,7 +408,7 @@ pre-commit reformat your files in next commit, and commit again.
 Quick dirty recipe:
 
 ```bash
-copier --force update
+copier update --trust
 git add .
 pre-commit run
 git add .


pedrobaeza:
I have just tried with copier 9.4.1 and the flag `-f` is still there.

bosd:
Indeed it is still there in the code.
I was getting the error on the same system, where I encountered this: https://github.com/Tecnativa/doodba-copier-template/issues/502#

Even in copier 9.2.0 the flag should be available.
I guess the copier installation was somehow corrupted.

And my bad. The comit I'm referencing is actually adding these flags..
Guess I was a bit :sleeping: after long debug session.

Closing this one now.

PR #513, opened 2024-12-26
[FIX] Better bus support out of the box

yajo:
This follows https://github.com/odoo/odoo/pull/126808 and hardens websockets out of the box.

@moduon MT-8520


pedrobaeza:
@yajo I think this should be put also for test.yaml, isn't it?

yajo:
Indeed, changed.

pedrobaeza (approved):

giarve:
I found that this is not working with the copier template as-is (traefik v2 template +letsencrypt). Is there anything that should be added in odoo.conf for this to work? 

Otherwise, it breaks webchat by default on prod and test using odoo 16.0

PR #515, opened 2025-01-21
Using unique service db names

josep-tecnativa:
CC @Tecnativa TT54598

When multiple Odoo projects are deployed within the same Docker network, there is a risk that an instance of Odoo may inadvertently connect to the database of another project due to identical service names. This PR addresses the issue by ensuring unique service names, preventing unintended cross-project database connections and ensuring proper project isolation.

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -1,4 +1,6 @@
 {% import "_macros.jinja" as macros -%}
+{%- import "_traefik2_labels.yml.jinja" as traefik2_labels -%}
+
 version: "2.4"
 
 services:
@@ -39,6 +41,7 @@ services:
     environment:
       POSTGRES_DB: *dbname
       POSTGRES_USER: *dbuser
+      PGHOST: "${DB_HOST:-db}"
       CONF_EXTRA: |
         work_mem = 512MB
     volumes:
--- a/devel.yaml.jinja
+++ b/devel.yaml.jinja
@@ -1,4 +1,6 @@
 {%- import "_macros.jinja" as macros -%}
+{%- import "_traefik2_labels.yml.jinja" as traefik2_labels -%}
+{%- set _key = traefik2_labels.key(project_name, odoo_version, "devel") -%}
 {% set whitelisted_hosts = (
   "cdnjs.cloudflare.com",
   "fonts.googleapis.com",
@@ -52,11 +54,12 @@ services:
       WDB_WEB_PORT: "{{ macros.version_major(odoo_version) }}984"
       # To avoid installing demo data export DOODBA_WITHOUT_DEMO=all
       WITHOUT_DEMO: "${DOODBA_WITHOUT_DEMO-false}"
+      PGHOST: {{ _key }}-db
     volumes:
       - ./odoo/custom:/opt/odoo/custom:ro,z
       - ./odoo/auto:/opt/odoo/auto:rw,z
     depends_on:
-      - db
+      - {{ _key }}-db
       {% for host in whitelisted_hosts -%}
       - proxy_{{ host|replace(".", "_") }}
       {% endfor -%}
@@ -77,13 +80,14 @@ services:
       {%- endif %}
 
   {% if postgres_version -%}
-  db:
+  {{ _key }}-db:
     extends:
       file: common.yaml
       service: db
     environment:
       POSTGRES_DB: *dbname
       POSTGRES_PASSWORD: odoopassword
+      DB_HOST: {{ _key }}-db
   {%- endif %}
 
   pgweb:
@@ -92,9 +96,10 @@ services:
     ports:
       - "127.0.0.1:{{ macros.version_major(odoo_version) }}081:8081"
     environment:
-      DATABASE_URL: postgres://{{ postgres_username }}:odoopassword@db:5432/devel?sslmode=disable
+      DATABASE_URL: postgres://{{ postgres_username }}:odoopassword@{{ _key }}-db:5432/devel?sslmode=disable
+      PGHOST: {{ _key }}-db
     depends_on:
-      - db
+      - {{ _key }}-db
 
   smtp:
     extends:
--- a/prod.yaml.jinja
+++ b/prod.yaml.jinja
@@ -30,8 +30,9 @@ services:
       {%- if smtp_relay_host %}
       SMTP_SERVER: smtplocal
       {%- endif %}
+      PGHOST: {{ _key }}-db
     depends_on:
-      - db
+      - {{ _key }}-db
       {%- if smtp_relay_host %}
       - smtp
       {%- endif %}
@@ -83,10 +84,12 @@ services:
       {%- endif %}
 
   {% if postgres_version -%}
-  db:
+  {{ _key }}-db:
     extends:
       file: common.yaml
       service: db
+    environment:
+      - DB_HOST={{ _key }}-db
     env_file:
       - .docker/db-creation.env
     restart: unless-stopped
@@ -136,8 +139,10 @@ services:
       - .docker/backup.env
       - .docker/db-access.env
     restart: unless-stopped
+    environment:
+      PGHOST: {{ _key }}-db
     depends_on:
-      - db
+      - {{ _key }}-db
       {%- if smtp_relay_host %}
       - smtp
       {%- endif %}
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -26,10 +26,42 @@ except ImportError:
 
 PROJECT_ROOT = Path(__file__).parent.absolute()
 SRC_PATH = PROJECT_ROOT / "odoo" / "custom" / "src"
+
+# _key = os.environ.get("_key", "").strip()
+# DB_SERVICE = os.environ.get("DB_HOST") or (_key + "-db" if _key else "db")
+
+
+def get_db_service_name():
+    """
+    Return the database service name that ends with '-db'. If not found,
+    fall back to any service containing 'postgres' or 'db'. As a last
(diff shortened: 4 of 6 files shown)


josep-tecnativa:
@ap-wtioit How does this look to you?

ap-wtioit:
> When multiple Odoo projects are deployed within the same Docker network, there is a risk that an instance of Odoo may inadvertently connect to the database of another project due to identical service names.

This to me seems not just a problem with db service but with all services (e.g. smtp). I'm not sure if it is wise to run multiple Odoo projects/instances in the same network. (Or are you talking about the inverseproxy_shared network only?)

What i get from the changes:
* DB_HOST is used to make sure that the odoo instance is connecting to a postgres service available under `{{_key}}-db`

When using this please make sure the value of _key is a valid hostname otherwise some key values might cause connection issues. E.g. `_` is sometimes used in service names (and can be resolved by docker dns accessible on 127.0.0.11 inside a container within a bridge network) but a correct DNS resolvable hostname would not contain an `_`. Otherwise i fear that this might lead to issues later (if odoo, postgres or another tool begin to validate db_name / PG_HOST other than resolving it)

We are using a modified prod.yaml with and additional .docker/db-host.env file where PGHOST is defined and on deployment the file is created by symlinking to something like `/path/to/postgres/containers/postgres_16.env`. I think this should still be compatible with the proposed change.

To prevent multiple `odoo` (we do not have the problem with db as we are using one postgres container per DB_VERSION) from resolving to the same name in shared networks we usually use aliases and do not change the service name in docker compose:
```yaml
services:
  odoo:
    ...
    networks:
      default: {}
      globalwhitelist_shared: {}
      inverseproxy_shared:
        aliases:
          - "${DB_NAME:-odoo}-${DOODBA_ENVIRONMENT:-prod}"
...
```
We use this for nginx providing access to Odoo from the web (instead of using traeffik).

So to sum it up, it looks ok. Only if someone already added PGHOST in db-access.env (or in our case new db-host.env) this could be a breaking change if `environment:` inside docker-compose.yaml (prod.yaml) wins over env files.

josep-tecnativa:
> > When multiple Odoo projects are deployed within the same Docker network, there is a risk that an instance of Odoo may inadvertently connect to the database of another project due to identical service names.
> 
> This to me seems not just a problem with db service but with all services (e.g. smtp). I'm not sure if it is wise to run multiple Odoo projects/instances in the same network. (Or are you talking about the inverseproxy_shared network only?)
> 
> What i get from the changes:
> 
>     * DB_HOST is used to make sure that the odoo instance is connecting to a postgres service available under `{{_key}}-db`
> 
> 
> When using this please make sure the value of _key is a valid hostname otherwise some key values might cause connection issues. E.g. `_` is sometimes used in service names (and can be resolved by docker dns accessible on 127.0.0.11 inside a container within a bridge network) but a correct DNS resolvable hostname would not contain an `_`. Otherwise i fear that this might lead to issues later (if odoo, postgres or another tool begin to validate db_name / PG_HOST other than resolving it)
> 
> We are using a modified prod.yaml with and additional .docker/db-host.env file where PGHOST is defined and on deployment the file is created by symlinking to something like `/path/to/postgres/containers/postgres_16.env`. I think this should still be compatible with the proposed change.
> 
> To prevent multiple `odoo` (we do not have the problem with db as we are using one postgres container per DB_VERSION) from resolving to the same name in shared networks we usually use aliases and do not change the service name in docker compose:
> 
> ```yaml
> services:
>   odoo:
>     ...
>     networks:
>       default: {}
>       globalwhitelist_shared: {}
>       inverseproxy_shared:
>         aliases:
>           - "${DB_NAME:-odoo}-${DOODBA_ENVIRONMENT:-prod}"
> ...
> ```
> 
> We use this for nginx providing access to Odoo from the web (instead of using traeffik).
> 
> So to sum it up, it looks ok. Only if someone already added PGHOST in db-access.env (or in our case new db-host.env) this could be a breaking change if `environment:` inside docker-compose.yaml (prod.yaml) wins over env files.
> 
> Edit: expected to need longer to find this. But indeed `enviroment:` overwrites `env_file:`: https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/#simple-example

Hi Andreas,

First of all, thank you for your detailed feedback and suggestions! 😊

To clarify, yes, this issue occurs specifically with the inverseproxy_shared network. It's a very odd problem because it only happens on a couple of servers, while on the rest everything works just fine.

Also, regarding your setup, we actually use one PostgreSQL container per Odoo instance, rather than having one container per PostgreSQL version. This approach helps us avoid conflicts between instances, but the shared network issue seems to still persist in those specific cases (I think it cloud be a kind of bug on Docker Compose).

Our PGHOST is currently defined [here](https://github.com/Tecnativa/doodba/blob/master/17.0.Dockerfile#L224). However, I need it to change dynamically based on the unique db service name, which is the reason behind my changes.

ap-wtioit (approved):

josep-tecnativa:
Could you please, check this out @pedrobaeza ? I think it's ready to go.

PabloEForgeFlow:
Hey @josep-tecnativa, it seems I'm a bit late to the party, but I just noticed this change after we ran an update on one of our projects and saw that scripts were failing due to the `db` service not existing anymore.

I personally believe that compose service names (together with volume and network names) should be treated as part of the template's "API" and should remain stable throughout (perhaps with the exception of major releases).

I've tried to replicate in various ways the issue this PR address and haven't been able to. You mentioned the issue only appears in a couple of servers, and just like @ap-wtioit said this does not appear to be a problem exclusive to the database service. Perhaps you could provide additional information on the specific setups that cause the issue so that we could address the compose issue instead?

In the meantime, do you think this same fix could be achieved by using network aliases instead of outright changing the default name of all database services? That way the fix would keep working while preventing potential disruptions to user scripts depending on the name of the database service.

Tardo:
For me, these changes break databases... :(

pedrobaeza:
For the record, this change doesn't break any database. If you do `copier update` on your scaffolding, your deployment tools may be affected, but nothing more.

The problem we faced was real, and reproduced already in 2 servers, having `docker compose` mixing the DBs from other versions of the same scaffolding (demo and prod for example). The conditions to get it are not clear though.

The options you have (apart from forking or freeze the copier updates, which I think are not adequate) are:

- Adapt your tooling to this change. We needed to do it, but they have been very few changes.
- Propose an alternative for having unique names for not having that mix. The network thing may work, but it seems everything gets more convoluted.

Tardo:
If I'm the only one with these problems... it must be because of how we update... What happened to me is that I got a PANIC: could not locate a valid checkpoint record.
I had to restore a copy, since I have not been able to repair the checkpoint.

pedrobaeza:
This is something only at docker compose level and how the DB service is called, so the checkpoint you are talking about should be another thing, and more if you didn't even update the scaffolding.

Tardo:
I have to investigate more what has happened to us... :/ 
It happened with the latest changes, but since we did not "attack" this repository, I was able to revert these changes and update again.

If I find anything that I think may be relevant to here I will comment :)

PabloEForgeFlow:
>     * Adapt your tooling to this change. We needed to do it, but they have been very few changes.

I guess we will be doing that for the time being, it's true that there are not many changes involved.
 
>     * Propose an alternative for having unique names for not having that mix. The network thing may work, but it seems everything gets more convoluted.

This is the change I would propose in the `db` service within `common.yaml`:
```
  db:
    image: ghcr.io/tecnativa/postgres-autoconf:{{ postgres_version }}-alpine
    shm_size: 4gb
    networks:
      default:
        aliases:
          - "{{ _key }}-db"
    environment:
      POSTGRES_DB: *dbname
      POSTGRES_USER: *dbuser
      PGHOST: "${DB_HOST:-db}"
      CONF_EXTRA: |
        work_mem = 512MB
    volumes:
      - db:/var/lib/postgresql/data:z
```

I believe this is both cleaner in terms of changes required and the impact it would have on external tooling. Odoo should still have a unique way to address the database (`{{ _key }}-db`), although if the same access is required externally you would need an additional alias in the `inverseproxy_shared` network.

pedrobaeza:
The last part is the problem, as the global inverse proxy shouldn't be touched in that cases.

PabloEForgeFlow:
I meant that you would need to set an alias in the `inverseproxy_shared` network, but not modify the proxy compose file itself. So, something like the following would go in `prod.yaml`:
```
db:
    extends:
      file: common.yaml
      service: db
    environment:
      - DB_HOST={{ _key }}-db
    env_file:
      - .docker/db-creation.env
    restart: unless-stopped
    {%- if postgres_exposed %}
    {%- if traefik_version == 3 %}
    networks:
      default:
      inverseproxy_shared:
        aliases:
          - "{{ _key }}-db"
    labels:
      traefik.enable: "true"
      traefik.docker.network: "inverseproxy_shared"
      {{- traefik3_labels_2.database(
        domains_prod,
        postgres_cidr_whitelist,
        _key,
        postgres_exposed_port,
        project_name,
      ) }}
    {%- else %}
    ports:
      - "{{ postgres_exposed_port }}:5432"
    {%- endif %}
    {%- endif %}

Aliases are defined on the service, but its effects are seen on all containers that belong to the network.

pedrobaeza:
> If I'm the only one with these problems... it must be because of how we update... What happened to me is that I got a PANIC: could not locate a valid checkpoint record. I had to restore a copy, since I have not been able to repair the checkpoint.

@Tardo I think I know what happened to you. You didn't do `down --remove-orphans`, so both DB containers continue to work in parallel, confusing the Odoo docker instance until data were screwed up.

PR #522, opened 2025-02-26
[IMP] Use network alias as an alternative to db service unique name

josep-tecnativa:
Proposed by @PabloEForgeFlow 
CC @Tecnativa TT54598

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -1,6 +1,4 @@
 {% import "_macros.jinja" as macros -%}
-{%- import "_traefik2_labels.yml.jinja" as traefik2_labels -%}
-
 version: "2.4"
 
 services:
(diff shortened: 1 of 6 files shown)


PabloEForgeFlow:
Hey, @josep-tecnativa thanks for the PR! We've tested the changes on one of our projects and everything is working as expected, so from our side it looks good.

josep-tecnativa:
Hi @PabloEForgeFlow ,
I think I've figured out the origin of the problem. When, with Traefik 3, we enable external access to the database with an external user, the db service is included in the Traefik network. Since this is a shared network, the Odoo containers from other instances might be connecting to their own db service or to the database that is in the inverseproxy_shared network.

Do you think adding an alias in the default network would be enough if we force Odoo to connect to that alias? I'm going to run some tests also to make sure it works.

PabloEForgeFlow:
That makes sense, docker probably chooses a network at random when two or more can resolve the same hostname and sometimes ends up with a different db on the shared network.

Yes, I believe doing so is enough, as the only possible DNS resolution of the new hostname would be the db service on the default network.

josep-tecnativa:
Tested and working. LGTM @pedrobaeza

pedrobaeza (approved):

PR #526, opened 2025-03-05
fix: allow odoo to find Postgres in production (‼️)

yajo:
If you deploy Odoo with `postgres_exposed=false`, your production server will go down without this fix. Odoo won't be able to find Postgres.

@moduon MT-9318

### DIFF
--- a/prod.yaml.jinja
+++ b/prod.yaml.jinja
@@ -93,12 +93,12 @@ services:
     env_file:
       - .docker/db-creation.env
     restart: unless-stopped
-    {%- if postgres_exposed %}
-    {%- if traefik_version == 3 %}
     networks:
       default:
         aliases:
             - "{{ _key }}-db"
+    {%- if postgres_exposed %}
+    {%- if traefik_version == 3 %}
       inverseproxy_shared:
     labels:
       traefik.enable: "true"


yajo:
This fix is pretty critical.

pedrobaeza:
The CI seems to still fails.

pedrobaeza:
Uhm, I have restarted it and it has failed in the first attempt.

yajo:
Ok now I understood what you mean. You're right, I pushed the fix.

prod.yaml.jinja
```diff
@@ -93,12 +93,12 @@ services:
     env_file:
       - .docker/db-creation.env
     restart: unless-stopped
-    {%- if postgres_exposed %}
-    {%- if traefik_version == 3 %}
     networks:
       default:
         aliases:
-            - "{{ _key }}-db"
+          - "{{ _key }}-db"
```
pedrobaeza, line 99: Is it OK this change of indentation?

yajo: It was "wrong" before. I mean: YAML would eat it fine, but it was inconsistent with the rest of the file.

josep-tecnativa (approved):

PR #532, opened 2025-03-19
[FIX] WebSocket scheme mismatch for Odoo 16+ by rewriting X-Forwarded-Proto

josep-tecnativa:
This PR adds a middleware that rewrites X-Forwarded-Proto to https for Odoo 16+ WebSocket connections. It prevents the scheme mismatch (https vs. wss) that would otherwise trigger a public session under ODOO_BUS_PUBLIC_SAMESITE_WS=1.

cc @Tecnativa TT55574

### DIFF
--- a/_traefik2_labels.yml.jinja
+++ b/_traefik2_labels.yml.jinja
@@ -81,6 +81,11 @@
       {%- endif %}
 {%- endmacro %}
 
+{%- macro custom_ws_middleware(key) -%}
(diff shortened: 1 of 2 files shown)


chienandalu:
Maybe this solves the mystery here: https://github.com/odoo/odoo/pull/194160

carlos-lopez-tecnativa (approved):
Tested and working fine, thanks!

PR #533, opened 2025-03-25
[FIX] Some format fixes

josep-tecnativa:
- DB_HOST: to allow multiple environment variables
- Remove line break after debug level on smtpreal

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -63,9 +63,9 @@ services:
       - SYS_PTRACE
     environment:
       DEFAULT_RELAY_HOST: "[{{ smtp_relay_host }}]:{{ smtp_relay_port }}"
-      {% if smtp_relay_version|float < 11.0 -%}
+      {%- if smtp_relay_version|float < 11.0 %}
       DMS_DEBUG: 0
-      {% else %}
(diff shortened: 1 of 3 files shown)


common.yaml.jinja
```diff
@@ -63,9 +63,9 @@ services:
       - SYS_PTRACE
     environment:
       DEFAULT_RELAY_HOST: "[{{ smtp_relay_host }}]:{{ smtp_relay_port }}"
-      {% if smtp_relay_version|float < 11.0 -%}
+      {%- if smtp_relay_version|float < 11.0 %}
```
pedrobaeza, line 66: What this `-` means?

josep-tecnativa: I noticed Jinja was adding an extra blank line in one of the branches. The dash (`-`) in Jinja removes trailing or leading whitespace around that control block, preventing unwanted blank lines. By aligning the dashes consistently in `if`, `else`, and `endif`, we avoid extra newlines for both `< 11.0` and `≥ 11.0`. I’ve tested this, and it now generates the YAML without unwanted blank lines.

pedrobaeza (approved):

PR #538, opened 2025-04-30
[IMP] Noto: fonts for all languages

giarve:
https://github.com/Tecnativa/doodba/discussions/533


giarve:
> Isn't this replacing the font for all the builds? Is that advisable?

It seems by tests it is not (although I expected it), just wanted to see whether any build passed.

I only thoroughly tested it on 16.0 image but seems the biggest issue is going to be apt repo compatibility

giarve:
I'll put it on doodba repo versioned Dockerfiles, here apt dependencies would need to be a templated file based on version and stuff like that 

but also possible if you want

PR #540, opened 2025-05-22
[FIX] Update skip parameter description and improve logging for missing modules

carlos-lopez-tecnativa:
- `_logger.warn` is deprecated; use `_logger.warning` instead.
- The format string was incorrect, which caused a `TypeError: not enough arguments for format string.`
- When the module is not in the list of modules, trying to remove it results in a `ValueError: list.remove(x): x not in list`

You can reproduce the error by running the following command:
`invoke test --modules module1,module2,module3 --skip module2,module4`

@pedrobaeza @josep-tecnativa could you please review this?

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -767,7 +767,7 @@ def _get_module_list(
         "extra": "Test all extra addons. Default: False",
         "private": "Test all private addons. Default: False",
         "enterprise": "Test all enterprise addons. Default: False",
-        "skip": "List of addons to skip. Default: []",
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):

david-banon-tecnativa (approved):

pedrobaeza (approved):

bosd:
Didn't know this command is possible.
Maybe add it to the docs?
'invoke test --modules module1,module2,module3'

PR #542, opened 2025-06-16
[IMP] Add suppor for Pydevd/Pycharm

david-banon-tecnativa:
Scaffolding support for debugging with pydevd/pycharm.
Depends on https://github.com/Tecnativa/doodba/pull/665
How to test: clone/update a template with the option `--vcs-ref=dck-imp-pydev_debugger`
Example:
`copier update --vcs-ref=dck-imp-pydev_debugger --trust -f`
Change dockerfile ref to
`ghcr.io/tecnativa/doodba:${ODOO_VERSION}-pr-665-test-onbuild`

Add pydevd-pycharm to the image, you should use a version compatible with your pycharm version (you can see on the menu to create a remote debugging task)

I recommend adding it to an ignored dependencies file ignored (ex 0-pip-local.txt), because it shouldn't end up in prod and 2 different developers may have different pycharm versions.

Rebuild the image, run git aggregate and then start with --pycharm
```
invoke img-build git-aggregate
invoke start --pydevd

```
@tecnativa TT56757

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@
 /doodba.*.code-workspace
 /src
 .idea/*
+# Ignore local dependencies
+/odoo/custom/dependencies/*-local.*
 
 # Project-specific docker configurations
 /.docker/
--- a/template/devel.yaml.jinja
+++ b/template/devel.yaml.jinja
@@ -40,6 +40,7 @@ services:
       INITIAL_LANG: "{{ odoo_initial_lang }}"
       LIST_DB: "true"
       DEBUGPY_ENABLE: "${DOODBA_DEBUGPY_ENABLE:-0}"
+      PYDEVD_ENABLE: "${DOODBA_PYDEVD_ENABLE:-0}"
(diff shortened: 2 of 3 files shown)


david-banon-tecnativa:
IMO this can be merged alongside https://github.com/Tecnativa/doodba/pull/665, tested yesterday while github was down and it works properly.
Please review both PRs and test if you can

As a sidenote
Pycharm has added a new debugger that uses debugpy (same as in vscode), https://blog.jetbrains.com/pycharm/2026/03/what-s-new-in-pycharm-2026-1/
However, I haven't managed to get it working, kinda connects but just does nothing, so i think it needs some more ironing out before real usage

david-banon-tecnativa:
This will not be merged, we made the choice to use the newer debugpy DAP method added in 2026.1, as the issues i had have already been fixed, and doesn't require a pycharm pro subscription . A pr by one of my colleges will be open with the required changes in the near future.

PR #544, opened 2025-06-25
[IMP] Remove version in compose yaml files.

david-banon-tecnativa:
In compose files, the version key is obsolete and does nothing, this romves it to avoid the warning.
Official compose spec document: https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-top-level-element-obsolete
Issue about it: https://github.com/docker/compose/issues/11628

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -1,5 +1,4 @@
 {% import "_macros.jinja" as macros -%}
-version: "2.4"
 
 services:
   odoo:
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Not sure if there's still some `docker-compose` systems out there...

david-banon-tecnativa:
I'll close this PR as we don't want to leave those on compose v1 behind. Maybe a copier flag or some other system could be implemented to make the version key in compose optional and remove the warning but its not planned right now.

PR #550, opened 2025-07-11
feat(odoo_listdb_staging): explicit option for exposing DB list in staging only

yajo:
This can be helpful if you want to have a single production environment
while having multiple databases for testing in staging.

@moduon MT-10847

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -20,7 +20,6 @@ services:
       PGDATABASE: &dbname {{ postgres_dbname }}
       PGUSER: &dbuser "{{ postgres_username }}"
       PROXY_MODE: "{% if odoo_proxy %}true{% else %}false{% endif %}"
-      LIST_DB: "{{ odoo_listdb | tojson }}"
     tty: true
     volumes:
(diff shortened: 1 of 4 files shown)


pedrobaeza:
And what happens with devel?

pedrobaeza:
OK, thanks. Can you check the error in CI?

yajo:
The logs say:

> ERROR: for smtp  toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
> toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit

AFAICS you should authenticate your Docker Hub pulls. That needs to be done administratively by someone from @Tecnativa because it will probably involve work such as changing the CI configuration and adding some secret tokens here and there.

Probably it will work if you just retry it.

yajo:
CI looks happy after you retried. Could we merge then?

josep-tecnativa:
Seems to be fixed when retried. This PR looks good to me

josep-tecnativa (approved):

PR #551, opened 2025-07-11
fix(ruff): modernize rules

yajo:
Modern Ruff versions yield this warning when scanning a doodba project:

  ``` warning: The top-level linter settings are deprecated in favour of
  their counterparts in the `lint` section. Please update the following
  options in `.ruff.toml`: - 'isort' -> 'lint.isort' - 'mccabe' ->
  'lint.mccabe' - 'per-file-ignores' -> 'lint.per-file-ignores' ```

@moduon MT-10479

### DIFF
--- a/.ruff.toml.jinja
+++ b/.ruff.toml.jinja
@@ -15,16 +15,16 @@ exclude = ["setup/*"]
 [format]
 exclude = ["setup/*"]
 
-[per-file-ignores]
+[lint.per-file-ignores]
 "__init__.py" = ["F401", "I001"]  # ignore unused and unsorted imports in __init__.py
 "__manifest__.py" = ["B018"]  # useless expression
 
-[isort]
+[lint.isort]
 section-order = ["future", "standard-library", "third-party", "odoo", "odoo-addons", "first-party", "local-folder"]
 
-[isort.sections]
+[lint.isort.sections]
 "odoo" = ["odoo"]
 "odoo-addons" = ["odoo.addons"]
 
-[mccabe]
+[lint.mccabe]
 max-complexity = 16


pedrobaeza:
Is this compatible with old versions?

pedrobaeza:
Well, I think we should ensure the compatibility with both, as not everyone is updating regularly `ruff`, or if it's available for the used distribution.

josep-tecnativa:
According to what I’ve been investigating, I’ve come to this conclusion:

Based on the OCA repo template ([link](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja)), Ruff is pinned to `v0.1.3` up until Odoo 18.0 (And we have this configuration like OCA on doodba-copier-template). That version does not support the `[lint.*]` syntax yet, so the current configuration without the `lint.` prefix should still work fine there.  

From Odoo 19.0 onwards, Ruff is pinned to `v0.6.8`, which already supports `[lint.*]` and emits deprecation warnings if we keep using `[isort]`, `[mccabe]`, or `[per-file-ignores]` at the top level.  

So the safe approach would be:  
- Odoo < 19 → keep the legacy syntax.  
- Odoo ≥ 19 → migrate to the `[lint.*]` syntax.  

This way we stay aligned with OCA’s pinned versions and avoid breaking older repos while fixing the warnings in newer ones.

yajo:
IMHO the OCA approach is wrong, because stuff depending on pre-commit depends on dev machine's packages, not on whatever is running on Odoo. We can use latest ruff with older Odoo versions, and it should just work. Maybe we can disable some rules for some versions, but most of the time it will be just OK. Also, what will make the code change is updating the template, not updating Odoo.

That said, you guys can decide how to do it here.

pedrobaeza:
We can't raise `ruff` version for all the existing Odoo versions, as this will change the way auto-formatting is done, provoking red CIs on existing repositories using the previous scaffolding with the previous `ruff` version. OCA is doing the same for the exact same reason.

The solution is to have one configuration file for 18.0+ and the other for the previous versions. @josep-tecnativa please proceed with the changes. I close this one, as we can't write in it.

PR #554, opened 2025-07-15
[IMP] Allow disabling autoreload on start

david-banon-tecnativa:
Also  fixes color for pre-commit on updatepot


pedrobaeza:
Please only one commit per unrelated fix, and check CI.

tasks_downstream.py
```diff
@@ -537,14 +537,14 @@ def lint(c, verbose=False):
 
 
 @task()
-def start(c, detach=True, debugpy=False):
+def start(c, detach=True, debugpy=False, _reload=True):
```
carlos-lopez-tecnativa, line 540: Why does the variable name start with an underscore?
```suggestion
def start(c, detach=True, debugpy=False, reload=True):
```

carlos-lopez-tecnativa (changes requested):

josep-tecnativa (approved):

PR #557, opened 2025-07-18
[IMP] Reduce tests

josep-tecnativa:
Reduce CI job count for faster, more reliable builds.


david-banon-tecnativa (commented):
I think we should also just run tests sequentially, running them in parallel makes no sense if we are just going to run them sequentially after that, and i believe most errors come from parallelization, even tho on local they pass just fine.

pedrobaeza (approved):

PR #558, opened 2025-07-31
[IMP] Add allowed paths

etobella:
This might allow to index images like favicon.

We are still testing if this fixes this issue:

image

As you can see, the favicon is not processed properly. The same happens in Bing.

The only reason we found is that favicon is related to web/image/website/1/favicon and traefik adds the x-robot-tag noindex.

Right now it is promising as Google Search Console App detected the favicon (before it had the same image that we find in google search). However, as we have not managed to reindex the main page, we are not 100% sure that it fixes it.

:crossed_fingers:

### DIFF
--- a/_traefik1_paths_labels.yml.jinja
+++ b/_traefik1_paths_labels.yml.jinja
@@ -48,7 +48,7 @@
 {%- endmacro %}
 
 {#- Basic labels for a single router #}
-{%- macro router(domain_group, key, suffix, rule=none, service=none, middlewares=()) %}
+{%- macro router(domain_group, key, suffix, rule=none, service=none, middlewares=(), priority=none) %}
       traefik.http.routers.{{ key }}-{{ suffix }}-{{ domain_group.loop.index0 }}.rule:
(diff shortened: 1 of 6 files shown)


josep-tecnativa:
Could the equal priorities (10) on /websocket and catch-all routers cause websockets to be routed to 8069 instead of 8072?

I did not realize when i tested it, but in the projects with updated template, we are facing this error:

`RuntimeError: Couldn't bind the websocket. Is the connection opened on the evented port (8072)?`

etobella:
@josep-tecnativa Only happens in 18? Where is the error raised? Server or JS?

pedrobaeza:
Only in 18. In the server.

josep-tecnativa:
@etobella @pedrobaeza Checking the logs last week, it seemed to be a false positive. I haven’t seen this log again and the websocket appears to be working correctly.

The functionality introduced by this PR is working perfectly; it may just take a few days/weeks for the results to become visible, as it depends on Google’s reindexing.

PR #569, opened 2025-09-22
[FIX] traefik: raise /websocket router priority so it always wins over the main (root “/”) router

josep-tecnativa:
Previously, the /websocket router and the main (root /) router had the same priority, so Traefik’s tie-breaker sometimes matched the root router and sent requests to 8069—triggering:

`RuntimeError: Couldn't bind the websocket. Is the connection opened on the evented port (8072)?`

This sets a higher priority (20) on the /websocket router (the main/root router remains at 10), ensuring /websocket deterministically targets the evented service (8072). In short: more specific rules must have higher priority.rity.


josep-tecnativa:
> More value in priority means that it's evaluated before?

Yes, that's it!

Issue #568, opened 2025-09-20
Deprecated top-level linter settings


kpassapk:
## Ruff warning
```
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `.ruff.toml`:
  - 'isort' -> 'lint.isort'
  - 'mccabe' -> 'lint.mccabe'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
```

## To Reproduce

```
git commit -m "..."
```

(this triggers pre-commit)

**Affected versions**:

ruff-pre-commit v0.1.3

**Expected behavior** 

There should be no warning

PR #570, opened 2025-09-23
[FIX] Depracated ruff options

josep-tecnativa:
This PR fixes https://github.com/Tecnativa/doodba-copier-template/issues/568


pedrobaeza:
This is the same as #551

PR #576, opened 2025-11-06
[IMP] Allow changing port prefix on devel

david-banon-tecnativa:
Allows changing the devel port prefix by calling starti with -p {port} ex:
`start -p 28`
this will expose everything on ports 28xxx, 
If not specified, will remain as it was (18 for v18.0, 17 for 17.0.etc)

@Tecnativa

### DIFF
--- a/devel.yaml.jinja
+++ b/devel.yaml.jinja
@@ -18,9 +18,9 @@ services:
       default:
       public:
     ports:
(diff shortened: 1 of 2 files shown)


josep-tecnativa (approved):

josep-tecnativa:
Great idea!

PR #580, opened 2025-12-09
[REF] Let pydevd handle symlink mappings

yajo:
By adding `PYDEVD_RESOLVE_SYMLINKS=1` to devel environments, all path mappings can be redudced to one.

This simplifies the creation of code-workspace files and makes it not necessary to run `invoke develop` when new modules are added to be debugged.

This feature did not exist when this code was built:
- https://github.com/fabioz/PyDev.Debugger/commit/2316bbdfc0d9809b6c17dbceb1da011bba43b5bb
- https://github.com/microsoft/debugpy/issues/743

@moduon MT-10900

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -111,7 +111,6 @@ def _get_cwd_addon(file):
 
 def _scan_subrepos_and_add_path_mappings(
     cw_config,
-    debugpy_configuration,
     firefox_configuration,
     chrome_configuration,
 ):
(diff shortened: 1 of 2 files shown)


david-banon-tecnativa:
I don't use the .workspace file, but i agree that simplifying it is a good cause. Mind rebasing to main see if CI is still red @yajo ?

tasks_downstream.py
```diff
@@ -254,7 +224,12 @@ def write_code_workspace_file(c, cw_path=None):
         "name": "Attach Python debugger to running container",
         "type": "python",
         "request": "attach",
-        "pathMappings": [],
+        "pathMappings": [
+            {
+                "localRoot": "${workspaceFolder:%s}/odoo" % root_name,
```
david-banon-tecnativa:
```suggestion
                "localRoot": f"${{workspaceFolder:{root_name}}}/odoo",
```

david-banon-tecnativa (changes requested):

PR #585, opened 2025-12-24
[FIX] remove invalid z option from docker volumes

HekkiMelody:
The :z option is only valid for bind mounts, not docker volumes. So it does not do anything, and it leads to the following warning from compose

```
WARN[0000] mount of type `volume` should not define `bind` option
```

It's not immediately clear from the documentation:

https://docs.docker.com/reference/compose-file/services/#short-syntax-5

but you can find a similar fix in other similar projects, e.g. https://github.com/cookiecutter/cookiecutter-django/pull/3981

### DIFF
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -22,7 +22,7 @@ services:
       PROXY_MODE: "{% if odoo_proxy %}true{% else %}false{% endif %}"
     tty: true
     volumes:
-      - filestore:/var/lib/odoo:z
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Wasn't this something needed in previous versions of Docker?

josep-tecnativa:
The truth is that the :z option is only valid for bind mounts, so this change shouldn’t break anything. However, since this affects volume configuration, I’d like to run some tests before merging to make sure there are no side effects.

Issue #541, opened 2025-05-30
The `version` attribute is deprecated in .yaml files


pleirb:
More than a bug, it is an error that does not affect but that always appears in the console and that can be avoided if the line: version: "2.4" that appears in all yaml files is removed.
The message that comes out is:

`the attribute "version" is obsolete, it will be ignored, please remove it to avoid potential confusion`

To reproduce it, simply run any of the environments and see the output on the console.

Affects the latest versions of Docker Compose

PR #590, opened 2025-12-29
[FIX] Docker compose version is obsolete

SirPyTech:
Fix https://github.com/Tecnativa/doodba-copier-template/issues/541.

Without this change, the following warning is shown when executing a compose file:
> the attribute "version" is obsolete, it will be ignored, please remove it to avoid potential confusion

### DIFF
--- a/setup-devel.yaml.jinja
+++ b/setup-devel.yaml.jinja
@@ -8,8 +8,10 @@
 #
 #   git clean -ffd
 
+{% if compose_version == "v1" %}
 version: "2.4"
 
+{% endif %}
 services:
   odoo:
     {%- if odoo_oci_image %}
--- a/common.yaml.jinja
+++ b/common.yaml.jinja
@@ -1,6 +1,8 @@
 {% import "_macros.jinja" as macros -%}
+{% if compose_version == "v1" %}
 version: "2.4"
 
+{% endif %}
 services:
   odoo:
     {%- if odoo_oci_image %}
--- a/copier.yml
+++ b/copier.yml
@@ -138,6 +138,18 @@ traefik_version:
     v2.4: 2
     v3.0: 3
 
+compose_version:
+  type: str
+  help: >-
+    Indicate the Docker Compose version.
+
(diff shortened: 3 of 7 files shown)


pedrobaeza:
One thing we can do is to add a question for adding that or not, letting it by not adding it by default.

setup-devel.yaml.jinja
```diff
@@ -8,8 +8,10 @@
 #
 #   git clean -ffd
 
+{%- if compose_version == "v1"%}
```
HekkiMelody: **style** (non-blocking)


```suggestion
{%- if compose_version == "v1" %}
```
what do you think?

setup-devel.yaml.jinja
```diff
@@ -8,8 +8,10 @@
...
 version: "2.4"
 
+{%- endif %}
```
HekkiMelody: **style** (non-blocking): 

I've learned that the `-` at the beginning or end of a tag strip whitespace before or after the tag

> You can also strip whitespace in templates by hand. If you add a minus sign (-) to the start or end of a block (e.g. a [For](https://jinja.palletsprojects.com/en/stable/templates/#for-loop) tag), a comment, or a variable expression, the whitespaces before or after that block will be removed

https://jinja.palletsprojects.com/en/stable/templates/

I think in this case we don't want the `-` before the `endif` because it would remove the empty line after the `version` key that you rightly left. I didn't test it, but that's my understanding at least.

SirPyTech: I tested it by executing:
```sh
copier copy --defaults --data compose_version=v1 --vcs-ref deprecated_version --trust gh:PyTech-SRL/doodba-copier-template TestDoodba
```

and you are correct: it removed the empty line, but we want to keep it, so I removed all the dashes `-`.
Thanks!

SirPyTech (commented):
I really didn't pay too much attention to spaces and dashes `-`, thanks for the review!

david-banon-tecnativa (changes requested):
One request, but otherwise LGTM, Thanks for working on this.

PR #613, opened 2026-03-17
Allow to configure backup password

SirPyTech:
Allow to configure the password for backups in a question.
The variable `BACKEND_PASSWORD` is mentioned in duplicity's documentation https://duplicity.gitlab.io/stable/duplicity.1.html#environment-variables:

> **BACKEND_PASSWORD**, _FTP_PASSWORD (deprecated)_
> 
>  Supported by most backends which are password capable. More secure than setting it in the backend url (which might be readable in the operating systems process listing to other users on the same machine).

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -815,3 +815,18 @@ backup_passphrase:
     restore backups.
 
     Which will be your backups passphrase?
+
+backup_backend_password:
+  secret: true
+  type: str
+  default: example-backup-backend-password
+  when: *backup_present
+  help: >-
+    ⚠️ This password is critical for security, so keep it safe. You will need it to
+    restore backups.
+
+    This password is supported by most duplicity backends (like SFTP) which are password
+    capable. If your backend uses it, setting it here will safely store it in an
+    environment variable, that is more secure than setting it in the backend url.
+
+    Which will be your backups password?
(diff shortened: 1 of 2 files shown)


copier.yml
```diff
@@ -711,3 +711,14 @@ backup_passphrase:
     restore backups.
 
     Which will be your backups passphrase?
+
+backup_password:
```
HekkiMelody: **comment**: having `backup_passphrase` and `backup_password` could get confusing

`backup_passphrase` is always needed by duplicity to decrypt the backup.

`backup_passphase` is (usually, but not always depending on the setup) needed to access the sftp folder/s3 bucket/etc where the backup resides. Maybe `backup_backend_passphrase`? It's a little verbose but clearer IMO.

HekkiMelody (commented):
Thank you for the contribution, this would be very useful to us. I left a couple of comments, please let me know what you think!

SirPyTech (commented):
Applied the suggestions, thanks!
Let me know if it looks better now

HekkiMelody (approved):
Code review, LGTM

quirino95 (approved):
Code and functional review: LGTM!

david-banon-tecnativa (approved):

SirPyTech:
Thanks for merging :pray:

PR #620, opened 2026-03-31
[FIX] _override_docker_command version fallback for debug mode

Neitherkx:
When using debugpy (VSCode debug), _override_docker_command fails on modern Docker Compose v2+ where devel.yaml no longer contains the version key.

The else clause only set the "2.4" fallback when orig_file was None. If the file existed but lacked version, docker_compose_file_version remained undefined, breaking debug mode.

### DIFF
--- a/tasks_downstream.py
+++ b/tasks_downstream.py
@@ -68,11 +68,12 @@ _logger = getLogger(__name__)
 
 def _override_docker_command(service, command, file, orig_file=None):
     # Read config from main file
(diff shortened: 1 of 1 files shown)


david-banon-tecnativa:
I can't reproduce your issue, mind sharing the steps you are taking and any error message?
I agree that the logic in _override_docker_command function is a bit weird and maybe needs some updating, but you should get no errors if you skip the version param and are running compose v2.

PR #626, opened 2026-04-29
[IMP] Add configurations to run odoo and tests using debugpy in pycharm

CarlosRoca13:
cc @Tecnativa

ping @david-banon-tecnativa @carlosdauden @sergio-teruel 

Needs review:
- [ ] Console don't show logs
image

### DIFF
--- a/template/tasks_downstream.py
+++ b/template/tasks_downstream.py
@@ -8,11 +8,14 @@ Contains common helpers to develop using this child project.
 import json
 import operator
 import os
+import platform
+import re
 import shutil
 import stat
 import subprocess
 import tempfile
 import time
+import xml.etree.ElementTree as ET
 from datetime import datetime
 from glob import iglob
 from itertools import chain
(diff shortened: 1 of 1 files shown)


CarlosRoca13:
I think that the console feature is not related to this PR and we can merge it

sergio-teruel (approved):
Tested on real environment

david-banon-tecnativa (commented):
Indeed, console issue isn't related to this PR, tested with the following file in a regular python env:

```python
print("Before")
# debug step on next line
numeber = 2+4
print(numeber)
print("After")
```
run with
```bash
python -m debugpy --wait-for-client --listen localhost:6988
```
VScode shows console, but Pycharm does not.

david-banon-tecnativa (approved):
Tested and working Good Job

CarlosRoca13:
Tested with user that is not using pycharm and it works fine too.

@pedrobaeza can we merge this?

PR #627, opened 2026-05-06
[IMP] Sync oca submodule to latest commit

david-banon-tecnativa:
Updates OCA repo template to latest commit, and uses it to obtain pre-commit config versions.
Also, bump python used to test legacy versions in ci as pre-commit wasn't behaving.
TT58388


pedrobaeza:
Have you checked in a case if this works properly?

PR #629, opened 2026-05-12
[DCK] Add test default proxies

david-banon-tecnativa:
TT59918 @Tecnativa @pedrobaeza

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -77,6 +77,24 @@ _migrations:
         - db-filter-prefix-default
         - "{{ _copier_conf.dst_path }}"
         - "{{ _copier_conf.answers_file }}"
+  - version: v9.2.1
+    before:
+      - - invoke
+        - --search-root={{ _copier_conf.src_path }}
+        - --collection=migrations
+        - migrate-to-new-proxy
+        - "{{ _copier_conf.dst_path }}"
+        - "{{ _copier_conf.answers_file }}"
+  - version: v9.6.0
+    before:
+      - - invoke
+        - --search-root={{ _copier_conf.src_path }}
+        - --collection=migrations
+        - add-new-domains
+        - "{{ _copier_conf.dst_path }}"
+        - "{{ _copier_conf.answers_file }}"
+        - "9.6.0"
+        - "--varname=whitelisted_hosts_test"
 
 # Questions for the user
 odoo_version:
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   // Language-specific settings
   "[python]": {
-    "editor.defaultFormatter": "ms-python.python"
+    "editor.defaultFormatter": "charliermarsh.ruff"
```
pedrobaeza: Why this change?

copier.yml
```diff
@@ -402,7 +420,38 @@ domains_migration:
 whitelisted_hosts_test:
   type: yaml
   multiline: true
-  default: []
+  default:
+    - accounts.google.com
+    - api.openai.com
```
pedrobaeza: I'm not sure about this one.

pedrobaeza (commented):
No version change in this one?

copier.yml
david-banon-tecnativa: Removed, we can add it later if needed.

.vscode/settings.json
david-banon-tecnativa: Reverted, VSCode said "python extension can't format python files", but it must be a mistake with my setup.

david-banon-tecnativa: As a reply to top comment, versions are changed when making a github release, once this is merged I'll make v9.6.0

copier.yml
```diff
@@ -402,7 +420,37 @@ domains_migration:
...
+    - iap-services.odoo.com
+    - maps.googleapis.com
+    - media-api.odoo.com
+    - olg.api.odoo.com
+    - www.google.com
+    - www.googleapis.com
+    - www.gravatar.com
+    - undraw.co
+    - updates.maxmind.com
+    - www.ecb.europa.eu # For OCA currency_rate_update module
```
pedrobaeza, line 444: Do comments work here?

pedrobaeza (approved):

copier.yml
david-banon-tecnativa: Yes, in the template `copier.yaml` we can do comments but they won't show on the rendered project.
We can't do them in `.copier.answers.yml` on projects because it gets replaced every copy/update from the answers dict.
If copier used https://pypi.org/project/ruamel.yaml/ keeping comments could be supported, however i don't think its worth the effort.

PR #630, opened 2026-05-15
[IMP] Use yaml anchors in devel and test allowed proxies.

david-banon-tecnativa:
TT59918


pedrobaeza (approved):
It sounds good. To be tested in real, and merge.

david-banon-tecnativa:
Added test to ensure proxy actually works as expected. @pedrobaeza

pedrobaeza (approved):

PR #631, opened 2026-05-22
[FIX] pg_isready

CristianoMafraJunior:
The pg_isready healthcheck was missing the -d flag, causing it to probe a database named after the user (odoo) instead of the actual database (devel). This flooded the logs with FATAL: database "odoo" does not exist every 5 seconds. Adding -d $$POSTGRES_DB makes the check target the correct database

cc @pedrobaeza


carlos-lopez-tecnativa (approved):
Tested and working.

pedrobaeza (approved):

PR #636, opened 2026-06-17
[IMP] disable postgres JIT

Tardo:
Odoo does not benefit from JIT (normal usage), so it's best to disable it by default.

Enabled by default in 12: https://www.postgresql.org/docs/12/runtime-config-query.html#GUC-JIT
Disabled by default in 19: https://www.postgresql.org/docs/19/runtime-config-query.html#GUC-JIT

More info: https://www.cybrosys.com/research-and-development/postgres/how-postgresql-jit-affects-odoo-performance-benchmarks

### DIFF
--- a/template/common.yaml.jinja
+++ b/template/common.yaml.jinja
@@ -43,6 +43,9 @@ services:
       PGDATA: "/var/lib/postgresql/data"
       CONF_EXTRA: |
         work_mem = 512MB
+        {%- if postgres_version|int > 11 %}
+        jit = off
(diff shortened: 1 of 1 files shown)


david-banon-tecnativa:
I don't think the "workload" proposed in cybrosis's article (installing sale module) really qualifies as a benchmark. 
Are you already running any instances with this change and have seen any improvement?.
I'll run tests on some of our projects with and without JIT and check if the results do indeed improve.

PR #642, opened 2026-07-16
[IMP] Remove unused questions & ask multiline questions last.

david-banon-tecnativa:
The following questions and the code they generated were not being used anywhere, so we are going to remove them.

- whitelist_docker_project_devel
- whitelist_docker_project_test

Also, all multi line questions are now the last to be answer, so if you update, you can just alt+enter once they start coming,
@tecnativa


pedrobaeza (commented):
The diff is huge, but it's due to the reorder of the questions.

juancarlosonate-tecnativa (approved):

christian-ramos-tecnativa (approved):

PR #647, opened 2026-08-06
[REM] Remove doodba-scaffolding to copier template migration test.

david-banon-tecnativa:
Its been 6 years since the transition from the old project to the new one. 
This test is being troublesome in CI, i can't manage to make it pass on local and i don't think we gain any value from maintaining this test, so i'm removing it.
@pedrobaeza @tecnativa


pedrobaeza (approved):
Yes, it's acceptable

PR #656, opened 2026-08-28
[FIX] Use a common pg version reference table and deprecate v9.6

### LINKED ISSUE #655: ValueError for Odoo version <= 14 due to a space in Postgres version choice
## Describe the bug

When selecting an Odoo Version <= 14, the postgres version choice crashes due to a space in the default value.
`ValueError: Invalid choice for 'postgres_version': '16 ' is not in ['', '9.6', '10', '11', '12', '13', '14', '15', '16', '17', '18']`

## To Reproduce
**Affected versions**:
Odoo <= 14

1. Select Odoo Version <=14
2. Leave all other options at default values
3. Template crashes when loading Postgres default choice

**Expected behavior** 
The Postgres Version is defaultet to 16 (Odoo <= 14) or to 13 (Odoo <= 12)

**Additional context** 
Happens likely due to the spaces in this place:
https://github.com/Tecnativa/doodba-copier-template/blob/5f88db3e58b2f6329d046ae3076da15bcbd65273/copier.yml#L545-L546

david-banon-tecnativa:
We use a fake copier question to store the pg per odoo versions in a single place, it won't end up in the answers file as it has when: False.
Deprecate v9.6 as it reachedd EOL more than 4 years ago and the backup service doesn't support it.
Fixes #655 
@tecnativa @pedrobaeza @patrick-rademacher-mm

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -95,6 +95,39 @@ _migrations:
         - "{{ _copier_conf.answers_file }}"
         - "9.6.0"
         - "--varname=whitelisted_hosts_test"
+pg_per_odoo:
+  when: False
+  type: yaml
+  # Odoo 11-13 supports pg 9.6 but that version is not supported by the backup service
+  # and is necessary to be able to perform all tests
+  default:
+    11.0:
+      oldest: "10"
+      latest: "13"
+    12.0:
+      oldest: "10"
+      latest: "13"
+    13.0:
+      oldest: "10"
+      latest: "16"
+    14.0:
+      oldest: "10"
+      latest: "16"
+    15.0:
+      oldest: "10"
+      latest: "17"
+    16.0:
+      oldest: "12"
+      latest: "18"
+    17.0:
+      oldest: "12"
+      latest: "18"
+    18.0:
+      oldest: "12"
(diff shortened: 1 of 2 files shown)


copier.yml
```diff
@@ -95,6 +95,39 @@ _migrations:
...
         - "9.6.0"
         - "--varname=whitelisted_hosts_test"
+pg_per_odoo:
+  when: False
+  type: yaml
+  # Odoo 11-13 supports pg 9.6 but that version is not supported by the backup service
+  # and is necessary to be able to perform all tests
+  default:
+    11.0:
+      oldest: "10"
```
pedrobaeza, line 105: Why do we put `oldest` if not used?

david-banon-tecnativa: It is used in the validator to check the pg version is supported by the oddoo version. Ex if someone on odoo 15 pg 11 upgraddes to odoo 16 without upgrading pg, copier itself will raise the error that odoo 16 doesn't support  pg 11

pedrobaeza: OK then.

pedrobaeza (approved):

PR #658, opened 2026-09-03
[IMP] add Github Access Token Logic

etobella:
@david-banon-tecnativa @pedrobaeza This is what I did to secure everything.

A few things:

1- the token is seen in the logs because git-aggregate is not offuscating it. We might need to fix that.
2- I tested only with 19.

Feel free to take it and modify it

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -14,6 +14,7 @@ _skip_if_exists:
   - odoo/custom/src/private/*/
   - odoo/custom/src/repos.yaml
   - odoo/custom/ssh/*
+  - .docker/github_access.env
 
 _tasks:
   - invoke after-update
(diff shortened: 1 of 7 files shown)


pedrobaeza:
For now, it seems to work again unauthenticated (GitHub has rolled back?)

etobella:
Well, it doesn't harm to have this here and wait :sob:

pedrobaeza:
Of course!

david-banon-tecnativa:
I've been working on https://github.com/Tecnativa/doodba/pull/728 into adopting it into doodba core, useing a barebones credential manager that returns a token provided via env for github calls that doesn't require changing all urls. I'll upload here a pr for the DCT part soon.

manuel-florido:
@etobella 

I have something like this hardcoded in local, it works with 16.0 too.

etobella:
@david-banon-tecnativa If your works, it looks much cleaner

PR #659, opened 2026-09-03
[IMP] Allow passing a github token to authenticate github calls

david-banon-tecnativa:
Depends on https://github.com/Tecnativa/doodba/pull/728
TT64371

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ env:
   PRE_COMMIT_HOME: ${{ github.workspace }}/.cache.~/pre-commit
   PYTEST_ADDOPTS: ${{ github.event.inputs.pytest_addopts }}
   PYTHONIOENCODING: "UTF-8"
-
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

carlosdauden (approved):

pedrobaeza:
With this, I suppose that not having the variable filled works like before, isn't it?

david-banon-tecnativa:
Yes, if DOODBA_GITHUB_TOKEN isn't set github calls are unauthenticated

HekkiMelody:
Should the same variable be added to `setup-devel.yaml`?

(or, this could be a good chance to refactor `setup-devel.yaml` to inherit from `common.yaml`)

david-banon-tecnativa:
I think we should just part ways with setup-devel, i've merged it into devel.yaml in https://github.com/Tecnativa/doodba-copier-template/pull/660

PR #660, opened 2026-09-03
[FIX] Merge setup-devel into devel.yaml

david-banon-tecnativa:
Less maintenance burden and it inherits changes to devel.

TT64371


template/devel.yaml.jinja
```diff
@@ -82,6 +82,21 @@ services:
...
+    profile: devel-setup
+    extends: odoo
+    entrypoint: autoaggregate
+    environment:
+      DEPTH_DEFAULT: 100
+      # XXX Export these variables before running setup to own the files
+      UID: "${DOODBA_GITAGGREGATE_UID:-1000}"
+      GID: "${DOODBA_GITAGGREGATE_GID:-1000}"
+      UMASK: "$DOODBA_UMASK"
+      user: root
```
HekkiMelody: **issue**: `user` should not be inside `environment` right?

HekkiMelody (changes requested):
Thank you for this!

HekkiMelody (approved):
Code review, LGTM!

pedrobaeza:
No change in the repo doc is needed?

david-banon-tecnativa:
It isn't mentioned anywhere but on a comment and is only used during the `inv git-aggregate` command.

CarlosRoca13 (approved):

pedrobaeza:
You should change it also in https://github.com/Tecnativa/doodba#autoaggregate

PR #661, opened 2026-09-15
[FIX] common.yml: add default value to github token to avoid warning

david-banon-tecnativa:
This avoids the sometimes repetitive warning when the variable isn't set.


template/common.yaml.jinja
```diff
@@ -1,8 +1,8 @@
-{% import "macros/_macros.jinja" as macros -%}
-{% if compose_version == "v1" %}
+{%- import "macros/_macros.jinja" as macros -%}
```
pedrobaeza: Are these changes related?

david-banon-tecnativa: Leftovers from another change, removed.

template/common.yaml.jinja
```diff
@@ -17,13 +17,13 @@ services:
         ODOO_VERSION: "{{ macros.version_minor(odoo_version) }}"
         UID: "${UID:-1000}"
         GID: "${GID:-1000}"
-        GITHUB_TOKEN: "${DOODBA_GITHUB_TOKEN}"
+        GITHUB_TOKEN: "${DOODBA_GITHUB_TOKEN:- default value please change}"
```
pedrobaeza, line 20: I suppose the default value is a blank space and the rest is not passed. It's the blank space processed?

david-banon-tecnativa: No, the default value with this syntax is literally "default value please change"

david-banon-tecnativa: Fixed with a blank space as default value

pedrobaeza: And a blank space is correctly interpreted where used?

PR #665, opened 2026-09-22
[FIX] Fix migrating from very old template versions

david-banon-tecnativa:
Migration from very old projects to the newest version wasn't working, and python2 versions couldn't even be updated.


pedrobaeza (approved):
Trusting on the tests

